Issue #28, opened 2025-11-27
stdout_callback deprecation .


ghpZ54K8ZRwU62zGVSePPs97yAv9swuAY0mVDR4:
https://github.com/hyperledger/fabric-x/blob/9cc5eaf33ff935da96a829c6386c6c6ec4d1b625/samples/tokens/ansible/ansible.cfg#L9

has been replaced by result_format .

ref :

https://docs.ansible.com/projects/ansible/latest/collections/community/general/yaml_callback.html

PR #29, opened 2025-11-27
 [ gh-28 ] use result_format .

ghpZ54K8ZRwU62zGVSePPs97yAv9swuAY0mVDR4:
use result_format .


pasquale95:
I would completely remove the `yaml` callback for Ansible and rely solely on the default Ansible behaviour, which uses JSON.
This has been already addressed in #32.

PR #38, opened 2026-01-06
Added support for production grade Docker image

anushka159-a:
It has changes for Production Docker images for fabric-x-tools.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@
 ###########################################
 # Stage 1: Build image
 ###########################################
-FROM golang:1.24 AS builder
+FROM golang:1.25 AS builder
 
 # List of CLI tools to build
 ARG FABRICX_TOOLS="configtxgen cryptogen configtxlator fxconfig"


.dockerignore
```diff
@@ -0,0 +1,22 @@
+#
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
```
pasquale95: add newline here

tools/Dockerfile
```diff
@@ -4,25 +4,58 @@
...
+    chown -R 10001:0 /home/fabricx && \
+    chmod 0755 /home/fabricx
 
 # Copy only the built tools
 COPY --from=builder /tmp/bin/* /usr/local/bin/
 
-CMD ["bash"]
+# OCI metadata labels
+LABEL name="fabric-x-tools" \
+    maintainer="IBM Research ZRL Decentralized Trust Group" \
```
pasquale95: "IBM Research Decentralized Trust Group"

tools/Dockerfile
```diff
@@ -4,25 +4,58 @@
...
+    version="${VERSION}" \
+    description="Fabric-X CLI tools (configtxgen, cryptogen, configtxlator, fxconfig, armageddon) packaged in a UBI image" \
+    license="Apache-2.0" \
+    vendor="IBM Research"
```
pasquale95: "IBM"

Makefile
```diff
@@ -71,4 +70,14 @@ lint: FORCE
...
+
+###########################################
+# Docker image build
+###########################################
+
+# Build the fabric-x-tools image for the current machine platform.
+.PHONY: build-fabric-x-tools-image
+build-fabric-x-tools-image:
+	./tools/build_image.sh --tag docker.io/hyperledger/fabric-x-tools:$(IMAGE_TAG) -f ./tools/Dockerfile
+
```
pasquale95: This can be removed since we don't build the image using the Makefile in the pipeline.

tools/Dockerfile
```diff
@@ -4,25 +4,58 @@
...
-# Install binaries directly from GitHub into /tmp/bin
-ENV GOBIN=/tmp/bin
+WORKDIR /go/src/github.com/hyperledger/fabric-x-tools
+
+# Install Fabric-X tooling
+RUN go install github.com/hyperledger/fabric-x/tools/configtxgen@${FABRIC_X_VERSION} && \
+    go install github.com/hyperledger/fabric-x/tools/cryptogen@${FABRIC_X_VERSION} && \
+    go install github.com/hyperledger/fabric-x/tools/configtxlator@${FABRIC_X_VERSION} && \
+    go install github.com/hyperledger/fabric-x/tools/fxconfig@${FABRIC_X_VERSION} && \
+    go install github.com/hyperledger/fabric-x-orderer/cmd/armageddon@${ARMAGEDDON_VERSION}
```
pasquale95: I realized that `armageddon` is embedded into the `fabric-x-orderer` image. So it can be removed from here.
In this way we should have now here only tools which are within `fabric-x`. Therefore, i would now switch to `COPY . .` + `go build...` instead of `go install` for faster image build.
Please do also the trick with copying `go.mod` and `go.sum` before actually doing the build.

.github/workflows/build-image.yml
```diff
@@ -64,12 +64,12 @@ jobs:
...
         with:
-          context: ./tools
-          file: ./tools/Dockerfile
+          context: .
+          file: ./Dockerfile
           platforms: linux/amd64,linux/arm64,linux/s390x
           push: true
           build-args: |
-            FABRIC_X_VERSION=${{ github.ref_name }}
+            FABRIC_X_VERSION=${{ env.VERSION }}
```
pasquale95: this is wrong. `FABRIC_X_VERSION` is not used anymore. You need to change this to `VERSION`.

pasquale95 (changes requested):

Dockerfile
```diff
@@ -8,7 +8,7 @@
 ###########################################
 # Stage 1: Build image
 ###########################################
-FROM golang:1.25 AS builder
+FROM golang:1.24 AS builder
```
pasquale95: Why have you reverted this?

PR #41, opened 2026-01-21
Feature/production image

anushka159-a:
Added OCI metadata labels as per the new format


pasquale95 (changes requested):
You need to provide the new ARGS in the build process, otherwise the image build will fail in the pipeline.

PR #44, opened 2026-01-29
adjust creation/cleanup of docker network

pasquale95:
#### Type of change

#### Description

To create the crypto material, `hyperledger.fabricx` uses `cryptogen` and `configtxgen` with docker containers. Since we use `container_network` in the inventory at `all.vars` level, it also affects `cryptogen` and `configtxgen` containers.  Hence, we should create the `container_network` before running those containers.

Similarly, I've adjusted the container network removal at `make clean` level.


pasquale95:
@arner agreed. It's better your approach. Closing.

PR #63, opened 2026-03-10
testing(fxconfig): add TLS to integration tests

remo-lab:
#### Type of change
- Test update

#### Description

Adds TLS coverage to the fxconfig integration tests. Previously the tests only ran with `--insecure`. This change parameterizes the tests so they run in two modes:

- `none` (existing behavior)
- `tls` (server-side TLS)

Ephemeral certificates are generated during the test using `tlsgen`, and copied into the test-node container using the same layout expected by the committer (`/server-certs/` and `/client-certs/`). The fxconfig client config is updated to include the generated CA so secure gRPC connections can be established.

All existing scenarios are now executed under both modes without duplicating test logic.

#### Implementation notes

- Added `tlsModes := []string{"none", "tls"}` and parameterized the integration tests with `t.Run`
- Introduced helper logic in `helpers_test.go` to generate temporary TLS credentials and configure the container
- For `tls` mode the test container runs without `--insecure` and all `SC_*_TLS_MODE` env vars are set to `tls`
- fxconfig config generation now optionally includes a `tls:` section with `enabled: true` and the generated CA

#### Test plan

- `go vet ./tools/fxconfig/...`
- `go test ./tools/fxconfig/internal/...`
- Integration tests pass in `none` mode (regression check)
- Integration tests pass in `tls` mode validating secure gRPC connections

#### Related issues

Closes #62

### DIFF
--- a/tools/fxconfig/integration/helpers_test.go
+++ b/tools/fxconfig/integration/helpers_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
 
+	"github.com/hyperledger/fabric-x-common/common/crypto/tlsgen"
 	fmsp "github.com/hyperledger/fabric-x-common/msp"
 	"github.com/hyperledger/fabric-x/tools/fxconfig/internal/adapters"
 	"github.com/hyperledger/fabric-x/tools/fxconfig/internal/app"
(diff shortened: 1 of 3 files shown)


mbrandenburger (changes requested):
Thank you @remo-lab for creating this PR to add the missing integration tests for TLS.  I think this PR is great and we can get it in once we make the CI happy :)

Moreover, we should test `tls` and `mtls` mode. WDYT?

PR #76, opened 2026-04-02
fix(fxconfig): support meta namespace updates

Sukuna0007Abhi:
#### Type of change

- Bug fix
- Test update

#### Description

This fixes #61 by removing the hardcoded _meta namespace version in namespace create/update tx generation.

fxconfig now reads the current _meta version from query results and uses it when building the transaction, so namespace operations still work after config/meta updates.
If _meta is not present yet (fresh network), it falls back to version 0 so first-time namespace creation still works.

Also added/updated tests for dynamic meta version handling and the no-_meta bootstrap case.

#### Related issues

#61

### DIFF
--- a/tools/fxconfig/integration/single_org_test.go
+++ b/tools/fxconfig/integration/single_org_test.go
@@ -480,4 +480,33 @@ func TestSingleOrgScenarios(t *testing.T) {
 			require.Contains(ct, nss, Namespace{Name: ns, Version: 1})
 		}, eventuallyTimeout, eventuallyTick)
 	})
+
+	t.Run("create_after_meta_namespace_version_update", func(t *testing.T) {
+		t.Parallel()
+
+		sourceNs := "s9_meta_update_source"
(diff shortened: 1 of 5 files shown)


Sukuna0007Abhi:
Pls review @mbrandenburger @arner @pasquale95 @liran-funaro

mbrandenburger (changes requested):
Thank you @Sukuna0007Abhi for creating this PR! I believe we should cover namespace updates also via an integration case scenario where we submit a config update to change the meta namespace version. Could you add such a scenario?

Sukuna0007Abhi:
> Thank you @Sukuna0007Abhi for creating this PR! I believe we should cover namespace updates also via an integration case scenario where we submit a config update to change the meta namespace version. Could you add such a scenario?

Pls review @mbrandenburger I've (maybe the correct one) added a scenario

mbrandenburger:
Thank you @Sukuna0007Abhi for working on this. Can you please check that all the checks are passing.

PR #78, opened 2026-04-02
feat(fxconfig): add chaincode rwset adapter scaffold

Sukuna0007Abhi:
#### Type of change

- New feature
- Test update

#### Description

This PR adds a small adapter in fxconfig to convert chaincode-style RW sets into Fabric-X transaction namespaces.

Scope:

- converted one namespace RW set into TxNamespace
- builded a tx from multiple namespace RW sets
- basic validation for required fields (namespace, key)
- This is groundwork for chaincode migration exploration, not a full compatibility implementation.

Covered cases:

- happy path mapping
- empty namespace
- empty key
- empty rwset list

Tested with:

- go test ./tools/fxconfig/internal/transaction
- go build ./tools/fxconfig

### DIFF
--- a/tools/fxconfig/internal/transaction/chaincode_adapter.go
+++ b/tools/fxconfig/internal/transaction/chaincode_adapter.go
@@ -1,0 +1,76 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package transaction
+
+import (
+	"fmt"
+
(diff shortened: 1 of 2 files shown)


Sukuna0007Abhi:
Pls review @mbrandenburger @pasquale95

mbrandenburger:
Hi @Sukuna0007Abhi thank you for creating this PR - while I see value in such a transaction builder - I don't believe it is needed here in fxconfig. Please consider to propose this building block in https://github.com/hyperledger/fabric-x-sdk. Thank you

PR #82, opened 2026-04-05
fix(fxconfig): use printer in version command

### LINKED ISSUE #151: `fxconfig version` command bypasses the CLI Printer interface
Hey folks,

Noticed a bit of technical debt while reading through `tools/fxconfig/internal/cli/v1/version.go`. The `fxconfig version` command currently bypasses the standard `cliCtx.Printer` interface entirely and just dumps its output straight to `cmd.Printf()`. 

There's actually a lingering `// TODO: use printer` comment right in the code pointing this out!

Why this matters: since `NewVersionCommand` doesn't currently take the `CLIContext` that our other commands do, if we ever switch up the global output formatting (like routing output to a JSON serializer via the Printer), the version command will completely ignore it and break the unified output format.

**Expected behavior:**
We should pass the `*CLIContext` down from `root.go` into `NewVersionCommand()` and swap out all the manual `cmd.Printf` calls with proper `ctx.Printer.Print()` calls.

**Steps to resolve:**
1. Update `NewVersionCommand` signature to accept `CLIContext`.
2. Refactor the `Run` function (and `showLine` helper) to natively use `ctx.Printer`.
3. Update `version_test.go` to mock the context and assert against the printer buffer instead of using cobra's raw capture stream.

If nobody else is grabbing this, I'm happy to throw together a quick PR to knock this out!

FirePheonix:
- Improvement:
Replaced ```cmd.Printf``` in the version command with the shared ```ctx.Printer```, making output consistent with the rest of the CLI.

#### A TODO existed in ```version.go``` to use the printer instead of ```cmd.Printf``` directly. The version command was the only command bypassing ```ctx.Printer```, duplicating output logic that the rest of the CLI handles consistently.

This PR resolves the TODO by passing ```*CLIContext``` into ```NewVersionCommand``` and initializing its own printer via PersistentPreRunE. This keeps it fully independent of config loading, which version doesn't need. The showLine helper and its unused imports are also removed.

### DIFF
--- a/tools/fxconfig/internal/cli/v1/root_test.go
+++ b/tools/fxconfig/internal/cli/v1/root_test.go
@@ -59,7 +59,7 @@ func TestPersistentPreRunE_ViaConfigFlag(t *testing.T) {
 	rootCmd := NewRootCommand(cliCtx, func(_ *config.Config) (app.Application, error) {
 		return &testApp{}, nil
 	})
-	rootCmd.SetArgs([]string{"--config", configPath, "info"})
+	rootCmd.SetArgs([]string{"--config", configPath, "version"})
 
 	require.NoError(t, rootCmd.Execute())
 
@@ -85,7 +85,7 @@ func TestPersistentPreRunE_ViaProjectConfig(t *testing.T) { //nolint:paralleltes
 	rootCmd := NewRootCommand(cliCtx, func(_ *config.Config) (app.Application, error) {
 		return &testApp{}, nil
 	})
-	rootCmd.SetArgs([]string{"info"})
+	rootCmd.SetArgs([]string{"version"})
 
 	require.NoError(t, rootCmd.Execute())
 
--- a/tools/fxconfig/internal/cli/v1/root_test.go
+++ b/tools/fxconfig/internal/cli/v1/root_test.go
@@ -59,7 +59,7 @@ func TestPersistentPreRunE_ViaConfigFlag(t *testing.T) {
 	rootCmd := NewRootCommand(cliCtx, func(_ *config.Config) (app.Application, error) {
 		return &testApp{}, nil
 	})
-	rootCmd.SetArgs([]string{"--config", configPath, "version"})
+	rootCmd.SetArgs([]string{"--config", configPath, "info"})
 
 	require.NoError(t, rootCmd.Execute())
 
@@ -85,7 +85,7 @@ func TestPersistentPreRunE_ViaProjectConfig(t *testing.T) { //nolint:paralleltes
 	rootCmd := NewRootCommand(cliCtx, func(_ *config.Config) (app.Application, error) {
 		return &testApp{}, nil
 	})
-	rootCmd.SetArgs([]string{"version"})
+	rootCmd.SetArgs([]string{"info"})
 
 	require.NoError(t, rootCmd.Execute())
 
--- a/tools/fxconfig/internal/cli/v1/version.go
+++ b/tools/fxconfig/internal/cli/v1/version.go
@@ -32,7 +32,7 @@ func NewVersionCommand() *cobra.Command {
 			p := cliio.NewCLIPrinter(cmd.OutOrStdout(), cmd.ErrOrStderr(), cliio.FormatTable)
 			p.Print("fxconfig\n")
 			p.Print(fmt.Sprintf(" %-16s %s\n", "Version:", metadata.Version))
-			p.Print(fmt.Sprintf(" %-16s %s\n", "Go version:", runtime.Version()))
+			p.Print(fmt.Sprintf(" %-16s %s\n", "Go Version:", runtime.Version()))
 			p.Print(fmt.Sprintf(" %-16s %s\n", "Commit:", metadata.CommitSHA))
 			p.Print(fmt.Sprintf(" %-16s %s\n", "OS/Arch:", osArch))
 		},
--- a/tools/fxconfig/internal/cli/v1/version.go
+++ b/tools/fxconfig/internal/cli/v1/version.go
@@ -18,7 +18,7 @@ import (
 
 // NewVersionCommand returns a command that displays version information.
 // It shows the fxconfig version, Go version, commit SHA, and OS/architecture.
-func NewVersionCommand(ctx *CLIContext) *cobra.Command {
+func NewVersionCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "version",
 		Short: "Display version information",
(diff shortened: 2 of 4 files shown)


mbrandenburger (changes requested):
thank you @FirePheonix for creating this PR. Can you please check the unit test errors and the linter errors.

tools/fxconfig/internal/cli/v1/root_test.go
```diff
@@ -59,7 +59,7 @@ func TestPersistentPreRunE_ViaConfigFlag(t *testing.T) {
 	rootCmd := NewRootCommand(cliCtx, func(_ *config.Config) (app.Application, error) {
 		return &testApp{}, nil
 	})
-	rootCmd.SetArgs([]string{"--config", configPath, "version"})
+	rootCmd.SetArgs([]string{"--config", configPath, "info"})
```
mbrandenburger: Can you explain why this change is needed?

tools/fxconfig/internal/cli/v1/root_test.go
```diff
@@ -85,7 +85,7 @@ func TestPersistentPreRunE_ViaProjectConfig(t *testing.T) { //nolint:paralleltes
 	rootCmd := NewRootCommand(cliCtx, func(_ *config.Config) (app.Application, error) {
 		return &testApp{}, nil
 	})
-	rootCmd.SetArgs([]string{"version"})
+	rootCmd.SetArgs([]string{"info"})
```
mbrandenburger: same question as above

tools/fxconfig/internal/cli/v1/version_test.go
```diff
@@ -11,68 +11,39 @@ import (
...
-			expectedOutput: []string{"Usage:", "fxconfig version"},
-			expectError:    false,
-		},
-		{
-			name:           "version command with invalid flag",
-			args:           []string{"version", "--invalid"},
-			expectedOutput: []string{"unknown flag: --invalid"},
-			expectError:    true,
+			name:           "version command output",
+			expectedOutput: []string{"fxconfig", "Version:", "Go version:", "OS/Arch:"},
```
mbrandenburger: Can you elaborate why we need to remove these test cases?

mbrandenburger (changes requested):
Hi @FirePheonix - Thank you for working on this PR. The changes on `version.go` look good to me. In particular, instantiating the printer within `PersistentPreRunE` is a good one. 
However, the changes on the tests are not so clear to me - should we need to change the tests at all?

tools/fxconfig/internal/cli/v1/root_test.go
FirePheonix: hey, so Cobra skips a parent's ```PersistentPreRunE``` when the child has its own. 

Since ```version``` now has one, root's hook (which loads config) never runs for it. The tests check that root's hook ran, so they must use ```info```, which doesn't override it.

so like - before this PR, version had no ```PersistentPreRunE```. 
So when you ran version, the root's hook ran first (loaded config, set up printer, etc.).

After this PR, version has its own ```PersistentPreRunE``` (to create its printer without needing a config file). Now when version runs, Cobra calls version's hook and skips root's entirely.

tools/fxconfig/internal/cli/v1/version_test.go
FirePheonix: The new tests call ```cmd.Run()``` directly. 
Cobra intercepts ```--help``` and unknown flags before Run fires, so those cases are untestable with this approach. 

The new tests are basically simpler: create the version command directly, call ```cmd.Run()``` directly.

The ```--help``` and ```--invalid``` cases require the FULL COBRA DISPATCH pipeline to work, Cobra intercepts ```--help``` before RUN ever fires. And then flag parsing errors happen before Run too. 

Calling ```cmd.Run()``` directly bypasses all that. You can't test those scenarios with the new approach.
They also tested Cobra's own behavior, not ours, i think fine to drop? 

let me know if i'm wrong here? I'll happily correct it...

tools/fxconfig/internal/cli/v1/version.go
```diff
@@ -28,21 +27,19 @@ func NewVersionCommand() *cobra.Command {
...
+			ctx.Printer = cliio.NewCLIPrinter(cmd.OutOrStdout(), cmd.ErrOrStderr(), cliio.FormatTable)
+			return nil
+		},
+		Run: func(_ *cobra.Command, _ []string) {
+			osArch := fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
+			ctx.Printer.Print("fxconfig\n")
+			ctx.Printer.Print(fmt.Sprintf(" %-16s %s\n", "Version:", metadata.Version))
+			ctx.Printer.Print(fmt.Sprintf(" %-16s %s\n", "Go version:", runtime.Version()))
+			ctx.Printer.Print(fmt.Sprintf(" %-16s %s\n", "Commit:", metadata.CommitSHA))
+			ctx.Printer.Print(fmt.Sprintf(" %-16s %s\n", "OS/Arch:", osArch))
```
mbrandenburger:
```suggestion
		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
			ctx.Printer = cliio.NewCLIPrinter(cmd.OutOrStdout(), cmd.ErrOrStderr(), cliio.FormatTable)
			return nil
		},
		Run: func(_ *cobra.Command, _ []string) {
			osArch := fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
			p := cliio.NewCLIPrinter(cmd.OutOrStdout(), cmd.ErrOrStderr(), cliio.FormatTable)
			p.Print("fxconfig\n")
			p.Print(fmt.Sprintf(" %-16s %s\n", "Version:", metadata.Version))
			p.Print(fmt.Sprintf(" %-16s %s\n", "Go version:", runtime.Version()))
			p.Print(fmt.Sprintf(" %-16s %s\n", "Commit:", metadata.CommitSHA))
			p.Print(fmt.Sprintf(" %-16s %s\n", "OS/Arch:", osArch))
```

mbrandenburger: I've looked into this change again and I was thinking it should be enough to create a local cli printer and just use it directly. By adding the PreRunE and inject the printer through the `ctx`, we modify shared state - even though this is not an issue for our cli application - it seems unnecessary.  

If we would go with this approach, there is no need for the changes in `root.go`, `root_test.go`, and not even `version_test.go`, right? 

WDYT?

FirePheonix:
@mbrandenburger i think your suggestion is better, a local printer in ```Run``` is the simpler solution. I'll remove the ```PersistentPreRunE``` from ```version.go```, drop the ```*CLIContext``` parameter, and create the printer locally in ```Run```.

and yea we wouldn't need those changes in tests too.

FirePheonix:
@mbrandenburger check the latest commits and let's run unit and linters over this again?

tools/fxconfig/internal/cli/v1/version.go
FirePheonix: done in latest commit.

tools/fxconfig/internal/cli/v1/version.go
```diff
@@ -29,20 +28,15 @@ func NewVersionCommand() *cobra.Command {
...
-			cmd.Printf("fxconfig\n")
-			showLine(cmd, "Version", metadata.Version)
-			showLine(cmd, "Go version", runtime.Version())
-			showLine(cmd, "Commit", metadata.CommitSHA)
-			showLine(cmd, "OS/Arch", fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH))
+			osArch := fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
+			p := cliio.NewCLIPrinter(cmd.OutOrStdout(), cmd.ErrOrStderr(), cliio.FormatTable)
+			p.Print("fxconfig\n")
+			p.Print(fmt.Sprintf(" %-16s %s\n", "Version:", metadata.Version))
+			p.Print(fmt.Sprintf(" %-16s %s\n", "Go version:", runtime.Version()))
```
mbrandenburger:
```suggestion
			p.Print(fmt.Sprintf(" %-16s %s\n", "Go Version:", runtime.Version()))
```

tools/fxconfig/internal/cli/v1/version_test.go
```diff
@@ -27,7 +27,7 @@ func TestVersionCommand(t *testing.T) {
 		{
 			name:           "version command with no args",
 			args:           []string{"version"},
-			expectedOutput: []string{"fxconfig", "Version:", "Go Version:", "OS/Arch:"},
+			expectedOutput: []string{"fxconfig", "Version:", "Go version:", "OS/Arch:"},
```
mbrandenburger:
```suggestion
			expectedOutput: []string{"fxconfig", "Version:", "Go Version:", "OS/Arch:"},
```

mbrandenburger: I don't think that we need these changes on the `version_test.go`. Please remove :)

mbrandenburger (changes requested):
Thank you @FirePheonix for your updates. A small minor change and then this PR is ready.

FirePheonix:
AAAAAAH, I'm so sorry I did that small "v" over there.. i might've missed it.
just pushed changes in the latest commit.

FirePheonix:
hey, also, @mbrandenburger i was wondering, may I work on the bccsp-config todo?
I've been working on it for a week.. I added the whole module, may i create the PR for that?

mbrandenburger:
> hey, also, @mbrandenburger i was wondering, may I work on the bccsp-config todo? I've been working on it for a week.. I added the whole module, may i create the PR for that?

May I ask you to ask this question on the corresponding issue :) Otherwise - it's difficult to track who is working on what. :)

FirePheonix:
ah, i think that issue is already being worked upon.
please review just THIS pr for now.. thank you

mbrandenburger (changes requested):
Thank you @FirePheonix We are almost there ... just keep the comments in `version_test.go` ... there is no need to remove them.

FirePheonix:
> Thank you @FirePheonix We are almost there ... just keep the comments in `version_test.go` ... there is no need to remove them.

done sir :)

mbrandenburger (approved):
Thank you @FirePheonix

PR #83, opened 2026-04-06
samples/tokens: make CORS and bind address configurable

kallal79:
Made samples/tokens safer and configurable by replacing permissive CORS with env-driven allow-list CORS, adding configurable bind address, updating owner/issuer/endorser to use it, adding CORS/bind tests plus a default CORS check script, and documenting the new security settings in the sample README.


kallal79:
Hi @pasquale95 @arner  , could you please review  PR #83 ?

PR #91, opened 2026-04-08
Refactor server startup in `main.go` files to handle errors from ListenAndServe in token sample servers

karthik120710:
Fix for : hyperledger/fabric-x-samples#13 
ListenAndServe` was called in a bare goroutine, silently discarding the returned error
  - Wrapped it in an anonymous function to capture and handle the error properly
  - `http.ErrServerClosed` is excluded since it is expected during graceful shutdown

  ## Problem
  In the following files, the HTTP server was started like this:

  go s.ListenAndServe()

  If the server failed to start (e.g. port already in use), the error was silently
  dropped and the program continued running — appearing healthy but serving no requests.

  ## Affected Files
  - samples/tokens/issuer/main.go:55
  - samples/tokens/owner/main.go:55
  - samples/tokens/endorser/main.go:46

  ## Fix
  Replaced the bare goroutine with:

  go func() {
      if err := s.ListenAndServe(); err != nil && err != http.ErrServerClosed {
          log.Fatal(err)
      }
  }()
  
  which gracefully handle error .

### DIFF
--- a/samples/tokens/endorser/main.go
+++ b/samples/tokens/endorser/main.go
@@ -43,7 +43,11 @@ func main() {
 		Handler: h,
 		Addr:    net.JoinHostPort("0.0.0.0", *port),
 	}
-	go s.ListenAndServe()
+	go func() {
+		if err := s.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Fatal(err)
+		}
+	}()
 
 	// Stop
 	stop := make(chan os.Signal, 1)
--- a/samples/tokens/endorser/main.go
+++ b/samples/tokens/endorser/main.go
@@ -43,9 +43,10 @@ func main() {
 		Handler: h,
 		Addr:    net.JoinHostPort("0.0.0.0", *port),
 	}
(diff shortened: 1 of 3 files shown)


arner:
Thanks for your contribution! This is definitely an improvement over how it was done before.

May I suggest one additional step, that ensures graceful shutdown of the server?

```go
	serverErr := make(chan error, 1)
	go func() {
		if err := s.ListenAndServe(); err != nil && err != http.ErrServerClosed {
			serverErr <- err
		}
	}()

	// Stop
	stop := make(chan os.Signal, 1)
	signal.Notify(stop, syscall.SIGINT, syscall.SIGTERM)

	select {
	case <-stop:
	case err := <-serverErr:
		log.Printf("server error: %v", err)
	}
```

arner (changes requested):
See comment above.

karthik120710:
hi @arner can u review my pr

arner (approved):
LGTM

mbrandenburger (approved):

PR #93, opened 2026-04-09
feat(fxconfig): add --format flag to info command

### LINKED ISSUE #149: fxconfig info `--format` flag is advertised in help but missing in implementation
Hey folks,

Noticed a small bug in the CLI while testing out the `fxconfig info` command. 

If you look at the command help (`fxconfig info --help`), the docs say you can use the `--format env` flag to print out the config as environment variables instead of yaml:
```text
  # Show configuration as environment variables
  fxconfig info --format env
```
But if you actually try to run that exact command, Cobra throws an unknown flag error:
```

$ fxconfig info --format env
Error: unknown flag: --format
```
Looking at `tools/fxconfig/internal/cli/v1/info.go` around line 42, the `--format` flag is never actually wired up to the command. It just hardcodes` yaml.Marshal` no matter what. There's also a leftover comment saying `// TODO: add flag to show yaml/env config.`

Steps to reproduce:

Run `fxconfig info --format env`
Expected behavior: The command should parse the flag and export the config formatted as environment variables, as the help text advertises.

Actual behavior: Crashes with `unknown flag: --format` since the flag isn't defined.

FirePheonix:

#### Description

This PR resolves an existing TODO in the `info` command by adding a `--format` flag, allowing users to view the effective configuration in either `yaml` (default) or `env` format.

### DIFF
--- a/tools/fxconfig/internal/cli/v1/info.go
+++ b/tools/fxconfig/internal/cli/v1/info.go
@@ -13,7 +13,6 @@ import (
 
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
-
 )
 
 func NewInfoCommand(ctx *CLIContext) *cobra.Command {
@@ -48,13 +47,13 @@ Examples:
 				sort.Strings(env)
 				ctx.Printer.Print(strings.Join(env, "\n") + "\n")
 			case "yaml":
-				fallthrough
-			default:
 				out, err := yaml.Marshal(ctx.Config)
 				if err != nil {
 					return err
 				}
 				ctx.Printer.Print(string(out))
+			default:
+				return fmt.Errorf("invalid --format: %s (want yaml|env)", format)
 			}
 			return nil
 		},
--- a/tools/fxconfig/internal/cli/v1/info.go
+++ b/tools/fxconfig/internal/cli/v1/info.go
@@ -7,14 +7,17 @@ SPDX-License-Identifier: Apache-2.0
 package v1
 
 import (
+	"fmt"
+	"sort"
+	"strings"
+
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
+
 )
 
-// NewInfoCommand returns a command that displays the effective configuration.
-// The configuration is shown as YAML after applying all overrides from
-// flags, environment variables, and config files.
 func NewInfoCommand(ctx *CLIContext) *cobra.Command {
+	var format string
 	cmd := &cobra.Command{
 		Use:   "info",
 		Short: "Display effective configuration",
@@ -39,15 +42,65 @@ Examples:
   # Show configuration as environment variables
   fxconfig info --format env`,
 		RunE: func(_ *cobra.Command, _ []string) error {
-			// TODO: add flag to show yaml/env config
-			out, err := yaml.Marshal(ctx.Config)
-			if err != nil {
-				return err
+			switch format {
+			case "env":
+				env := toEnv("FXCONFIG", ctx.Config)
+				sort.Strings(env)
+				ctx.Printer.Print(strings.Join(env, "\n") + "\n")
+			case "yaml":
+				fallthrough
+			default:
+				out, err := yaml.Marshal(ctx.Config)
+				if err != nil {
+					return err
+				}
+				ctx.Printer.Print(string(out))
 			}
-			ctx.Printer.Print(string(out))
 			return nil
 		},
 	}
 
+	cmd.Flags().StringVar(&format, "format", "yaml", "Output format (yaml|env)")
+
 	return cmd
+}
+
+func toEnv(prefix string, cfg any) []string {
+	// we use yaml marshaling as a shortcut to get a map representation of the config
+	// that respects all yaml tags and omitempty.
+	out, err := yaml.Marshal(cfg)
+	if err != nil {
+		return nil
+	}
+
+	var m map[string]any
+	if err := yaml.Unmarshal(out, &m); err != nil {
+		return nil
+	}
+
+	return flatten(prefix, m)
+}
+
+func flatten(prefix string, m map[string]any) []string {
+	var result []string
+	for k, v := range m {
+		key := strings.ToUpper(strings.ReplaceAll(k, "-", "_"))
+		if prefix != "" {
+			key = prefix + "_" + key
+		}
+
+		switch val := v.(type) {
+		case map[string]any:
+			result = append(result, flatten(key, val)...)
+		case []any:
+			var strVals []string
+			for _, item := range val {
+				strVals = append(strVals, fmt.Sprintf("%v", item))
+			}
+			result = append(result, fmt.Sprintf("%s=%s", key, strings.Join(strVals, ",")))
+		default:
+			result = append(result, fmt.Sprintf("%s=%v", key, val))
+		}
+	}
+	return result
 }
--- a/tools/fxconfig/internal/cli/v1/info.go
+++ b/tools/fxconfig/internal/cli/v1/info.go
@@ -46,7 +46,10 @@ Examples:
 		RunE: func(_ *cobra.Command, _ []string) error {
 			switch format {
 			case "env":
-				env := toEnv("FXCONFIG", ctx.Config)
+				env, err := toEnv("FXCONFIG", ctx.Config)
+				if err != nil {
+					return err
+				}
 				slices.Sort(env)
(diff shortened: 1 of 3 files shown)


tools/fxconfig/internal/cli/v1/info.go
```diff
@@ -7,14 +7,17 @@ SPDX-License-Identifier: Apache-2.0
 package v1
 
 import (
+	"fmt"
+	"sort"
+	"strings"
+
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
+
```
mbrandenburger:
```suggestion
```

tools/fxconfig/internal/cli/v1/info.go
```diff
@@ -39,15 +42,65 @@ Examples:
...
+				sort.Strings(env)
+				ctx.Printer.Print(strings.Join(env, "\n") + "\n")
+			case "yaml":
+				fallthrough
+			default:
+				out, err := yaml.Marshal(ctx.Config)
+				if err != nil {
+					return err
+				}
+				ctx.Printer.Print(string(out))
```
mbrandenburger, line 60: I am wondering if that should be more explicit. For instance:

```go
		switch format {
		case "env":
			...
 		case "yaml":
			...
		default:
			return fmt.Errorf("invalid --format: %s (want yaml|env)", format)
		}
```

tools/fxconfig/internal/cli/v1/info_test.go
```diff
@@ -56,3 +56,31 @@ func TestInfoCommand_NilConfigPrintsNull(t *testing.T) {
...
+			MSP: config.MSPConfig{
+				LocalMspID: "TestOrg",
+			},
+		},
+		Printer: cliio.NewCLIPrinter(&outBuf, &outBuf, cliio.FormatTable),
+	}
+
+	cmd := NewInfoCommand(ctx)
+	err := cmd.Flags().Set("format", "env")
+	require.NoError(t, err)
```
mbrandenburger, line 122: We should also test invalid formats.

tools/fxconfig/internal/cli/v1/info_test.go
```diff
@@ -56,3 +56,31 @@ func TestInfoCommand_NilConfigPrintsNull(t *testing.T) {
...
+	var outBuf bytes.Buffer
+	ctx := &CLIContext{
+		Config: &config.Config{
+			Logging: config.LoggingConfig{
+				Level: "debug",
+			},
+			MSP: config.MSPConfig{
+				LocalMspID: "TestOrg",
+			},
+		},
```
mbrandenburger, line 116: I suggest we extend this test config to cover all aspects of flatten. For instance, we could use the configuration as in https://github.com/hyperledger/fabric-x/tree/main/tools/fxconfig#configuration-file-example

mbrandenburger (changes requested):
Thank you @FirePheonix for creating this PR. 

As part of this PR we also should to update the fxconfig documentation.

A few more comments below.

tools/fxconfig/internal/cli/v1/info.go
FirePheonix: done in latest commit.

tools/fxconfig/internal/cli/v1/info_test.go
FirePheonix: yea, you're very right. i'll make these changes.

tools/fxconfig/internal/cli/v1/info_test.go
FirePheonix: added in docs

FirePheonix:
@mbrandenburger i've done the requested changes.
please review again and let me know if improvements are required.

tools/fxconfig/internal/cli/v1/info_test.go
```diff
@@ -56,3 +57,101 @@ func TestInfoCommand_NilConfigPrintsNull(t *testing.T) {
...
+			},
+			MSP: config.MSPConfig{
+				LocalMspID: "Org1MSP",
+				ConfigPath: "/path/to/msp",
+			},
+			TLS: config.TLSConfig{
+				Enabled:        boolPtr(true),
+				ClientKeyPath:  "/path/to/client.key",
+				ClientCertPath: "/path/to/client.crt",
+				RootCertPaths:  []string{"/path/to/ca.crt"},
```
mbrandenburger, line 81: This is a good improvement; I think additionally we should make sure that also multiple certs are printed correctly. WDYT?

tools/fxconfig/internal/cli/v1/info.go
```diff
@@ -7,14 +7,16 @@ SPDX-License-Identifier: Apache-2.0
...
+	"slices"
+	"strings"
+
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
 )
 
-// NewInfoCommand returns a command that displays the effective configuration.
-// The configuration is shown as YAML after applying all overrides from
-// flags, environment variables, and config files.
```
mbrandenburger, line 16: The comment should be updated - not entirely removed :)

mbrandenburger (changes requested):
Thank you @FirePheonix for working on this! Great improvement; I have a few comments below.

tools/fxconfig/internal/cli/v1/info.go
FirePheonix: ahh, on it.
accidentally removed them.

tools/fxconfig/internal/cli/v1/info_test.go
FirePheonix: yea, you're right, lemme do..

tools/fxconfig/internal/cli/v1/info.go
```diff
@@ -39,15 +44,65 @@ Examples:
...
+	// that respects all yaml tags and omitempty.
+	out, err := yaml.Marshal(cfg)
+	if err != nil {
+		return nil
+	}
+
+	var m map[string]any
+	if err := yaml.Unmarshal(out, &m); err != nil {
+		return nil
+	}
```
mbrandenburger, line 84: Probably, not a real problem but we should return the error here instead of swallowing it and deal with it in `RunE()` above.

tools/fxconfig/internal/cli/v1/info_test.go
```diff
@@ -56,3 +57,130 @@ func TestInfoCommand_NilConfigPrintsNull(t *testing.T) {
...
+	require.Contains(t, output, "FXCONFIG_MSP_LOCALMSPID=Org1MSP")
+	require.Contains(t, output, "FXCONFIG_MSP_CONFIGPATH=/path/to/msp")
+	require.Contains(t, output, "FXCONFIG_TLS_ENABLED=true")
+	require.Contains(t, output, "FXCONFIG_TLS_CLIENTKEY=/path/to/client.key")
+	require.Contains(t, output, "FXCONFIG_TLS_CLIENTCERT=/path/to/client.crt")
+	require.Contains(t, output, "FXCONFIG_TLS_ROOTCERTS=/path/to/ca.crt")
+	require.Contains(t, output, "FXCONFIG_ORDERER_ADDRESS=localhost:7050")
+	require.Contains(t, output, "FXCONFIG_ORDERER_CHANNEL=mychannel")
+	require.Contains(t, output, "FXCONFIG_QUERIES_ADDRESS=localhost:7001")
+	require.Contains(t, output, "FXCONFIG_NOTIFICATIONS_ADDRESS=localhost:7001")
```
mbrandenburger, line 139: I think we should add another check that the timeouts (durations) are properly rendered in "human-readable" form. For instance, `ConnectionTimeout: 30 * time.Second` should be rendered as `30s`.

mbrandenburger (changes requested):
Thank you @FirePheonix. I think are almost there. A few minor changes below.

tools/fxconfig/internal/cli/v1/info_test.go
FirePheonix: done in the latest commit.

tools/fxconfig/internal/cli/v1/info.go
FirePheonix: hmmm, so, for the error handling: 
I made ```toEnv``` return ```([]string, error)``` so errors propagate up to ```RunE```
see latest commit, lemme know if any more changes

mbrandenburger (approved):
Thank you @FirePheonix - great effort!

PR #102, opened 2026-04-12
fix: remove panics in fsc.go

### LINKED ISSUE #101: fix: replace panic() calls with proper error returns in fsc.go
Problem
fsc.go contains 4 instances of panic() in production code paths. Using panic() causes the entire application to crash instead of allowing errors to be handled gracefully. Additionally, these panics discard the original underlying error context, making debugging and diagnosis more difficult.

Expected Fix
Replace all panic() calls in production code with proper error returns. Functions should propagate errors to the caller using Go’s standard error handling patterns.

Daniel-1600:
Type of change
- Bug fix

 Description
Replaced 4 instances of `panic()` in `fsc.go` with proper error returns.
Panics crash the entire application and discard the original error, making
failures hard to diagnose. Each panic is now replaced with
`return nil, fmt.Errorf("...: %w", err)` to propagate errors correctly.

#### Related issues
Closes #101

### DIFF
--- a/samples/tokens/owner/service/fsc.go
+++ b/samples/tokens/owner/service/fsc.go
@@ -59,7 +59,7 @@ var (
 func (f FabricSmartClient) Balances(ctx context.Context, wallet string) ([]Amount, error) {
 	mgmt, err := token.GetManagementService(f.node)
 	if err != nil {
-		panic("configuration error, check the logs")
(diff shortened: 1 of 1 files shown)


arner (approved):
Thanks for your contribution!

mbrandenburger (approved):

PR #110, opened 2026-04-14
fix: return non-zero exit code for failed tx submit with --wait

### LINKED ISSUE #96: `fxconfig tx submit --wait` returns success exit code even when transaction status is failure
### Description:
When `--wait` is used, the command prints the returned status string but always returns `nil` unless transport/API call itself fails. This means statuses such as `ABORTED_*`, `REJECTED_*`, or `MALFORMED_*` still produce process exit code `0`.

This is a functional bug because the command documentation explicitly describes success/failure behavior for waited submissions, and CI/automation cannot reliably detect failed transactions.

### Severity:
High

### Location:
`tools/fxconfig/internal/cli/v1/tx_submit.go` + `newTxSubmitCommand` (RunE)

### Steps to Reproduce (if possible):
1. Start a Fabric-X environment where a submitted tx can be deterministically rejected/aborted (e.g., duplicate txID or malformed tx payload).
2. Run `fxconfig tx submit <tx-file> --wait` and let the committer return a non-`COMMITTED` status.
3. Observe command output shows a failure status (for example `MALFORMED_*`), but shell exit code remains `0`.

### Expected Behavior:
For waited submissions, non-`COMMITTED` statuses should produce a non-zero process exit code (or equivalent explicit command failure).

### Actual Behavior:
The command returns success (`nil` error) for any status value if waiting completed without transport error.

### Root Cause Analysis:
`newTxSubmitCommand` calls `SubmitTransactionWithWait`, prints `committerpb.Status_name[...]`, and immediately returns `nil` without validating status semantics. The status enum confirms failure states are distinct from committed (`COMMITTED=1`, failures include `ABORTED_*`, `REJECTED_*`, `MALFORMED_*`).

### Suggested Fix (High-Level, no code):
After `SubmitTransactionWithWait`, explicitly check returned status and treat only `COMMITTED` as success. Return a structured error (or non-zero status) for all other enum values.

Test gap: existing tests in `tools/fxconfig/internal/cli/v1/tx_submit_test.go` cover wait success but do not assert failure-status exit behavior.

Logging/observability gap: status is only printed as text; there is no machine-readable failure output contract beyond exit code, making this bug especially impactful in automation.

aaradhychinche-alt:
## Summary
This PR fixes exit code behavior for fxconfig tx submit --wait so failed transaction outcomes return a non-zero exit code.

## Problem
The command printed the final transaction status but still exited with code 0 for failure statuses. This caused automation and CI workflows to treat failed transactions as successful.

## Root Cause
In the wait path, the command always returned nil after printing transaction status, regardless of the status value.

## Solution
The command now returns an error when the final status is not COMMITTED, and returns nil only for COMMITTED.
Status printing remains unchanged and still uses the enum name mapping.
Type casts are intentionally preserved because TxStatus is int while committerpb.Status is int32.

## Impact
- No change to successful transaction behavior
- CLI status output format is unchanged
- Failure cases now correctly produce a non-zero exit code

## Testing
- Updated the success wait test to assert committed status behavior
- Added a failure wait test using ABORTED_SIGNATURE_INVALID status
- Verified tx submit tests pass with expected success and failure semantics

fixes #96

### DIFF
--- a/tools/fxconfig/internal/cli/v1/tx_submit.go
+++ b/tools/fxconfig/internal/cli/v1/tx_submit.go
@@ -69,7 +69,10 @@ Examples:
 					fmt.Sprintf("Transaction status: %s", committerpb.Status_name[int32(status)]), //nolint:gosec
 				)
 				if status != int(committerpb.Status_COMMITTED) {
-					return fmt.Errorf("transaction failed with status: %s", committerpb.Status_name[int32(status)]) //nolint:gosec
+					return fmt.Errorf(
+						"transaction failed with status: %s",
+						committerpb.Status_name[int32(status)], //nolint:gosec
+					)
 				}
 				return nil
 			}
--- a/tools/fxconfig/internal/cli/v1/tx_submit.go
+++ b/tools/fxconfig/internal/cli/v1/tx_submit.go
@@ -68,6 +68,9 @@ Examples:
 				ctx.Printer.Print(
 					fmt.Sprintf("Transaction status: %s", committerpb.Status_name[int32(status)]), //nolint:gosec
 				)
+				if status != int(committerpb.Status_COMMITTED) {
+					return fmt.Errorf("transaction failed with status: %s", committerpb.Status_name[int32(status)]) //nolint:gosec
+				}
 				return nil
 			}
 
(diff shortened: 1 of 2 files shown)


mbrandenburger (changes requested):
Thank you @aaradhychinche-alt for creating this PR. Can you please look at the linter and make it happy. Thank you

aaradhychinche-alt:
> Thank you @aaradhychinche-alt for creating this PR. Can you please look at the linter and make it happy. Thank you

done with it !!  please take a look at it @mbrandenburger

mbrandenburger (approved):
Thank you @aaradhychinche-alt

Issue #108, opened 2026-04-13
gRPC clients fall back to plaintext when TLS is disabled - no certificate validation


MayankSharmaCSE:
## Summary

When `tls.enabled` is set to `false` (or left unset), the gRPC client in `fxconfig` silently downgrades to a completely **plaintext connection** with zero certificate validation. Because the default value of `enabled` is `false`, every fresh deployment that doesn't explicitly opt into TLS is running without any transport security out of the box.

## Where

**File:** `tools/fxconfig/internal/client/client.go`, lines 42-44  
**Function:** `createSecOpts`

```go
if !tlsConfig.IsEnabled() {
    return &secOpts, nil   // returns empty SecureOptions → UseTLS = false → plaintext
}
```
This returns a bare comm.SecureOptions{} struct - UseTLS stays false, no root CAs are loaded, and the resulting grpc.ClientConn is dialed without any transport credentials.

The IsEnabled() check lives in tools/fxconfig/internal/config/config.go:107-111 and defaults to false when the pointer is nil.

## Why this matters
1.No encryption in transit. Channel payloads (transaction proposals, block events, notification streams) are sent as cleartext over the network.
2.No server identity verification. Without root CA certificates, there is nothing stopping a man-in-the-middle from impersonating the orderer, query, or notification service.
3.Unsafe default. The enabled field defaults to false, which means operators who forget to (or don't know to) flip the flag are silently running insecure. This is especially risky in cloud or multi-tenant environments where network trust boundaries are not guaranteed.

## Steps to reproduce
1.Create or use a config file where `tls.enabled` is either explicitly `false` or simply omitted.
2.Start `fxconfig` — all three clients will connect over plaintext gRPC.
3.Inspect the wire traffic (e.g., with `tcpdump` or Wireshark) to confirm payloads are unencrypted.

## Suggested fix
A few options to consider (not mutually exclusive):

Default `enabled` to `true`. Flip the struct tag default from `"false"` to `"true"` so TLS is on unless an operator deliberately disables it. This is the "secure by default" approach.
Log a warning when TLS is off. If keeping false as the default is intentional for local dev, emit a clear warning at startup (e.g., `"TLS is disabled for <service> — connections are not encrypted"`) so operators are aware.
Require an explicit opt-out. Instead of silently accepting a missing `enabled` field, require the operator to set `enabled: false` explicitly if they truly want plaintext, and error on `nil`.

MayankSharmaCSE:
[...]

PR #114, opened 2026-04-14
fix(fxconfig): use system CAs when TLS is enabled but no root certs provided

MayankSharmaCSE:
### Problem

When `tls.enabled` is set to `false` or left unset, the gRPC client in `fxconfig` silently downgrades to a plaintext connection with zero certificate validation. Since `enabled` defaults to `false`, fresh deployments run without transport security out of the box.

Issue: [#108](https://github.com/hyperledger/fabric-x/issues/108)

### Solution

This PR addresses the security concern by ensuring that when TLS is enabled but no root certificates are provided, the client falls back to using the system's default CA pool for certificate validation instead of falling back to plaintext.

**Changes made:**

1. Modified `createSecOpts()` in `client.go` - only set custom `ServerRootCAs` when explicit root cert paths are provided
2. When `RootCertPaths` is empty or nil, `ServerRootCAs` is set to `nil` (not an empty slice), allowing Go's TLS stack to use the system default CA pool
3. Added unit tests to verify the system CA fallback behavior

### Security Impact

- Eliminates silent plaintext fallback when TLS is enabled but no custom root certs configured
- Fresh deployments with TLS enabled now properly validate server certificates using system CAs
- No more plaintext connections when TLS is intended to be used

---

### Files Changed

- `tools/fxconfig/internal/client/client.go`
- `tools/fxconfig/internal/client/client_test.go`
- `tools/fxconfig/internal/client/notifications.go`

### Test Cases Added

- `TestCreateSecOpts_TLS_EmptyRootCertPaths_UsesSystemCAs`
- `TestCreateSecOpts_TLS_NilRootCertPaths_UsesSystemCAs`

---

### DIFF
--- a/tools/fxconfig/internal/client/client.go
+++ b/tools/fxconfig/internal/client/client.go
@@ -46,16 +46,19 @@ func createSecOpts(tlsConfig *config.TLSConfig) (*comm.SecureOptions, error) {
 	secOpts.UseTLS = true
 	secOpts.ServerNameOverride = tlsConfig.ServerNameOverride
 
-	// set rootCAs
-	serverRootCAs := make([][]byte, 0, len(tlsConfig.RootCertPaths))
-	for _, rootCertPath := range tlsConfig.RootCertPaths {
-		rootCert, err := loadFile(rootCertPath)
-		if err != nil {
-			return nil, err
+	// set rootCAs — only if custom roots were configured; otherwise let Go use
+	// the system default CA pool so certificate validation is still active.
+	if len(tlsConfig.RootCertPaths) > 0 {
+		serverRootCAs := make([][]byte, 0, len(tlsConfig.RootCertPaths))
+		for _, rootCertPath := range tlsConfig.RootCertPaths {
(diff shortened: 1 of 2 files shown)


mbrandenburger:
@MayankSharmaCSE Thank you for this PR - I don't think that this PR resolves the issue you are reporting in #108

MayankSharmaCSE:
@mbrandenburger,my bad,I mistakenly mixed up the issues and pushed commits meant for a different task. I'll clean this up right away and make sure the PR aligns with the intended fix.

MayankSharmaCSE:
@mbrandenburger, i have successfully cleaned up the branch history.
The branch now has a single, clean commit (754553a). Could you please review it.

mbrandenburger:
@MayankSharmaCSE thank you for updating this PR. I'm still struggling to understand how this PR addresses the issue reported in #108. The core problem reported in #108 is that we currently use `tls.enabled=false` as default if not specified via the config file. This problem is not addressed in this PR. 

I suggest to close this PR and we continue the discussion in #108 and create a new PR.

PR #119, opened 2026-04-16
fix(tokens): fail fast on curl errors in demo test script

Prachi194agrawal:
#### Type of change

- Bug fix

#### Description

`samples/tokens/scripts/test.sh` used `set +e` at the script entry point, which
instructs bash to ignore all non-zero exit codes. Combined with `trap exit 1 INT`
(which only handles Ctrl+C), every `curl -f` call inside `run_test()` could fail
silently — the script would continue executing subsequent steps and exit with code `0`,
falsely reporting the demo as successful.

**Example:** if the issuer's `POST /issuer/issue` call fails (e.g. container not ready,
HTTP 500), bash ignores the failure and proceeds to the `transfer` and `transactions`
calls, which then operate on a non-existent token. `make test` would pass in CI even
when the network is broken.

**Changes made:**
- Replace `set +e` with `set -e` — script now exits immediately on any command failure
- Replace `trap exit 1 INT` with `trap cleanup INT ERR` — network teardown is triggered
  on both Ctrl+C and any unexpected error, not just keyboard interrupt
- Add `trap - INT ERR` inside `cleanup()` — prevents recursive trap firing if
  `stop_network` itself fails during cleanup

#### Additional details

Verified with `bash -n` (syntax check passes). Live terminal proof of before/after:

```bash
# Before (set +e): curl failure silently swallowed
set +e; curl -f -sS http://127.0.0.1:1 >/dev/null; echo "continued=yes"
# Output: curl_exit=7, continued=yes  ← BUG

# After (set -e): script halts on failure
bash -c 'set -e; curl -f -sS http://127.0.0.1:1 >/dev/null; echo "continued=yes"'
# Output: subshell_exit=7  ← "continued=yes" never printed, correct
```

#### Related issues

Fixes hyperledger/fabric-x-samples#6

### DIFF
--- a/samples/tokens/fabric3.mk
+++ b/samples/tokens/fabric3.mk
@@ -38,9 +38,10 @@ clean-fabric:
 # Start the targeted hosts (e.g. make fabric-fabric start).
 .PHONY: start-fabric
 start-fabric:
-	@"$(FABRIC_SAMPLES)/test-network/network.sh" down || true
-	@$(CONTAINER_CLI) rm -f peer0org1_token_namespace_ccaas peer0org2_token_namespace_ccaas >/dev/null 2>&1 || true
-	@$(CONTAINER_CLI) network inspect fabric_test >/dev/null 2>&1 && $(CONTAINER_CLI) network rm fabric_test || true
+	@if $(CONTAINER_CLI) network inspect fabric_test >/dev/null 2>&1; then \
+		echo "Error: existing fabric_test network detected. Run 'make teardown' first."; \
+		exit 1; \
+	fi
 	"$(FABRIC_SAMPLES)/test-network/network.sh" up createChannel -i 3.1.1
 	INIT_REQUIRED="--init-required" "$(FABRIC_SAMPLES)/test-network/network.sh" deployCCAAS  -ccn token_namespace -ccp "$(abspath $$CONF_ROOT)/namespace" -cci "init"
 	./scripts/cp_fabric3.sh
--- a/samples/tokens/fabric3.mk
+++ b/samples/tokens/fabric3.mk
@@ -38,6 +38,9 @@ clean-fabric:
 # Start the targeted hosts (e.g. make fabric-fabric start).
 .PHONY: start-fabric
 start-fabric:
+	@"$(FABRIC_SAMPLES)/test-network/network.sh" down || true
+	@$(CONTAINER_CLI) rm -f peer0org1_token_namespace_ccaas peer0org2_token_namespace_ccaas >/dev/null 2>&1 || true
+	@$(CONTAINER_CLI) network inspect fabric_test >/dev/null 2>&1 && $(CONTAINER_CLI) network rm fabric_test || true
 	"$(FABRIC_SAMPLES)/test-network/network.sh" up createChannel -i 3.1.1
 	INIT_REQUIRED="--init-required" "$(FABRIC_SAMPLES)/test-network/network.sh" deployCCAAS  -ccn token_namespace -ccp "$(abspath $$CONF_ROOT)/namespace" -cci "init"
 	./scripts/cp_fabric3.sh
--- a/samples/tokens/scripts/test.sh
+++ b/samples/tokens/scripts/test.sh
@@ -17,7 +17,12 @@ function print_section_header() {
 
 ## Cleanup and stop network on abort
 function cleanup() {
+    local exit_code=$?
+    local failed_cmd="$BASH_COMMAND"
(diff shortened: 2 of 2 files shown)


samples/tokens/scripts/test.sh
```diff
@@ -61,9 +63,9 @@ function run_test() {
 }
 
 # Script Start
-set +e
+set -e
```
arner: Can you please change this to `set -eE`? That way the ERR is also inherited by bash functions.

samples/tokens/scripts/test.sh
```diff
@@ -17,6 +17,8 @@ function print_section_header() {
 
 ## Cleanup and stop network on abort
 function cleanup() {
+    trap - INT ERR
```
arner, line 22: The reason the script fails gets drowned out by the logs of the network stopping. How about this?

```bash
function cleanup() {
    local exit_code=$?
    local failed_cmd="$BASH_COMMAND"
    trap - INT ERR
    set +e
    stop_network
    echo "Error: command '$failed_cmd' exited with status $exit_code" >&2
    exit 1
}
```

arner (changes requested):

samples/tokens/scripts/test.sh
Prachi194agrawal: Applied your suggestion now capturing $BASH_COMMAND and $exit_code before teardown so the failure reason is visible in the logs.

samples/tokens/scripts/test.sh
Prachi194agrawal: Done! Changed to set -eE so the ERR trap is inherited by bash functions.

arner:
Very nice. Can you please do two more things?

1. Rebase on latest `main`
2. Change line 77 to `if [[ "$PLATFORM" == "fabricx" || "$PLATFORM" == "xdev" ]]; then`

The rebase should hopefully fix the build issue you reported as well. Thanks!

Prachi194agrawal:
Sir, @arner 
done changes as recommended but some checks are still failing like unit test and fabric3

arner:
That's strange, it works when I try it locally... Could you run the tests locally and see if they succeed for you or not? You probably have to first `export PLATFORM=fabric3` before running the script.

Prachi194agrawal:
[Screencast from 2026-04-24 17-46-44.webm](https://github.com/user-attachments/assets/55a61b43-d2ae-498c-a418-c0ba0400497f)
@arner, can you review it again? I made many changes. Can you check if they are wrong or not?

samples/tokens/scripts/test.sh
```diff
@@ -39,37 +44,96 @@ function init_fabricx() {
...
+        wait_until_ready "endorser" "http://localhost:9300/readyz"
+    fi
+
+    wait_until_ready "issuer" "http://localhost:9100/readyz"
+    wait_until_ready "owner1" "http://localhost:9500/readyz"
+    wait_until_ready "owner2" "http://localhost:9600/readyz"
+}
+
+## Run curl with retries for transient startup errors
+function curl_with_retry() {
```
arner, line 77: Do you think maybe we can skip this in favour of built-in curl features?

https://everything.curl.dev/usingcurl/downloads/retry.html

Something like:
```
curl -fS ... \
  --retry "${CURL_MAX_ATTEMPTS:-6}" \
  --retry-delay "${CURL_RETRY_SLEEP_SECONDS:-2}" \
  --retry-all-errors
```

Same for the wait_until_ready() function; would be great if we can reduce it; even if we miss some features of the current version, simplicity is worth a lot!

samples/tokens/fabric3.mk
```diff
@@ -38,6 +38,9 @@ clean-fabric:
 # Start the targeted hosts (e.g. make fabric-fabric start).
 .PHONY: start-fabric
 start-fabric:
+	@"$(FABRIC_SAMPLES)/test-network/network.sh" down || true
```
arner: Hopefully this is not necessary; it should be stopped before starting a new one, and we might not want to do this without explicit agreement by the user. Did you find this necessary some time? A lighter solution could be to do some kind check to see if it's already running (port open, docker with grep?) and then exit 1.

arner (changes requested):
Great that you got it working! And the retries are a good approach. Maybe it can still work with less code - left some suggestions.

Prachi194agrawal:
It is failing again after the changes suggested, @arner.

iliecirciumaru:
Hi @Prachi194agrawal 

PR #123, opened 2026-04-16
fix(fxconfig): clean up subscriber on context cancellation

### LINKED ISSUE #109: fxconfig: Subscribe leaks stale subscriber on canceled context

cynox-66:
### Problem

`Subscribe()` adds a receiver channel to the `subscribers` map before checking whether the context is still valid.

If the context is already canceled (or gets canceled while waiting to enqueue the request), the function returns an error but leaves the channel in the map.

Subsequent calls for the same `txID` see an existing subscriber and assume an upstream request is already active, so they skip sending a new request. This leads to the caller hanging indefinitely.

---

### Fix

* Release the mutex before any blocking operations
* Add rollback logic to remove the subscriber on context cancellation
* Delete the map entry if no subscribers remain

---

### Notes

* The rollback removes only the channel created by the current call (using pointer equality)
* No locks are held during blocking operations

---

### Testing

* Existing tests pass, including `TestNotificationClient_Subscribe_ContextCanceled`

---

Closes #109

### DIFF
--- a/tools/fxconfig/internal/client/notifications.go
+++ b/tools/fxconfig/internal/client/notifications.go
@@ -80,14 +80,33 @@ func (n *NotificationClient) Subscribe(ctx context.Context, txID string) (chan i
 	receiverCh := make(chan int, 1)
 
 	n.subscribersMu.Lock()
-	defer n.subscribersMu.Unlock()
-
 	subscribers := n.subscribers[txID]
 	n.subscribers[txID] = append(subscribers, receiverCh)
+	isFirst := len(subscribers) == 0
+	n.subscribersMu.Unlock()
 
-	if len(subscribers) > 0 {
+	if !isFirst {
 		// we already have an active subscription for this txID
 		return receiverCh, nil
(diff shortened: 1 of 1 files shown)


mbrandenburger:
Thank you @cynox-66 for creating this PR. As in #109 @rootp1 is assigned to this task. However, your solution is really nice. I am wondering if we can cherry pick your changes and integrate them in #169. WDYT?

cynox-66:
Thanks, that makes sense!

Happy to have the relevant parts cherry-picked into #169. I took a look at the latest updates there — the targeted removal approach looks solid.

One small thought: since `removeSubscriber` can run alongside the dispatcher in `listen()`, it might be worth a brief note about that interaction (even though the mutex + guard handles it safely).

Happy to help review or contribute further if needed — this part of the notification flow is quite interesting to dig into.

mbrandenburger:
Thank you @cynox-66  Closing this one as we fix the issue in #169

PR #136, opened 2026-04-16
fix: make Makefile cross-platform for Windows native terminals

### LINKED ISSUE #133: make tools natively fails on Windows due to Unix-specific mkdir/rm commands
## Summary
The [Makefile](cci:7://file:///c:/Users/Windows/Desktop/fabric-x/Makefile:0:0-0:0) relies on Unix-specific shell commands like `mkdir -p` and `rm -rf`. When a developer tries to build the project locally on a native Windows terminal, the build fails because these commands aren't recognized.

## Expected Behaviour
Running `make tools` on a Windows machine (via PowerShell or CMD) should compile the tools into the `bin` directory without throwing shell command errors.

## Actual Behaviour
The build aborts immediately when executing the `mkdir -p` instruction inside the `$(BUILD_DIR)/%` build target because `-p` is not a valid flag for the Windows `mkdir` command.

## Reproduce
1. Clone the repository on a Windows machine.
2. Open native native PowerShell or CMD.
3. Ensure GNU Make is installed.
4. Run `make tools`.
5. See the build fail at the `mkdir -p bin` step.

## Proposed Fix
Update the [Makefile](cci:7://file:///c:/Users/Windows/Desktop/fabric-x/Makefile:0:0-0:0) to be cross-platform. We can either use Go's native `go build -o` (which automatically handles directory creation in modern Go) and bypass the raw `mkdir` call, or we can use OS detection in the Makefile to map `$(MKDIR)` to the correct command based on the operating system.

Hashan25-Dev:
## Summary
This PR fixes #133 by enabling native Windows environments to execute `make tools` and `make clean` out of the box, without requiring Git Bash or MSYS2.

## Details
Previously, the [Makefile](cci:7://file:///c:/Users/Windows/Desktop/fabric-x/Makefile:0:0-0:0) relied on inline Unix-specific shell commands like `@mkdir -p`, `rm -rf`, and inline environment variable assignments (`GOBIN=... cmd`). These naturally fail on Windows CMD or PowerShell. 

I've abstracted these into cross-platform macros at the top of the [Makefile](cci:7://file:///c:/Users/Windows/Desktop/fabric-x/Makefile:0:0-0:0):
- `$(call MKDIR,...)`
- `$(call RM,...)`
- `$(call TOUCH,...)`
- `$(call SET_ENV,...)`

On Linux/macOS, these evaluate to the exact same Unix flags `mkdir -p`, `rm -rf`, `touch`, and inline variables as before, so this does not alter or break the existing CI/Linux environments. 
On Windows, they expand into their CMD-compatible equivalents (e.g. `if not exist...`, `rmdir /s /q`).

## Verification
- Tested `make clean` and `make tools` natively on Windows and confirmed all 4 binaries build successfully into the `bin/` dir. 
- Leaves existing Linux workflows 100% untouched.

Fixes #133

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,19 @@
 .SUFFIXES:
 MAKEFLAGS += --no-builtin-rules
 
+# Cross-platform shell command abstraction
+ifeq ($(OS),Windows_NT)
+    MKDIR   = if not exist $(subst /,\,$1) mkdir $(subst /,\,$1)
+    RM      = if exist $(subst /,\,$(1)) rmdir /s /q $(subst /,\,$(1))
+    TOUCH   = type nul > $(subst /,\,$1)
+    SET_ENV = set $1=$2 &&
+else
+    MKDIR   = mkdir -p $1
+    RM      = rm -rf $1
+    TOUCH   = touch $1
+    SET_ENV = $1=$2
+endif
+
 BUILD_DIR ?= bin
 
 PKGNAME = github.com/hyperledger/fabric-x-common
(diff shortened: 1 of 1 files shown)


mbrandenburger:
Thank you @Hashan25-Dev for creating this PR! I think we should not move this effort forward to make the Makefile cross-platform compatible. This will open a rabbit-hole of additional work and testing.

PR #138, opened 2026-04-16
  fix(fxconfig)!: enable TLS by default to prevent plaintext fallback vulnerability

### LINKED ISSUE #108: gRPC clients fall back to plaintext when TLS is disabled - no certificate validation

MayankSharmaCSE:
## Summary

  Enable TLS by default across all service configurations (orderer, queries, notifications). Previously TLS was disabled by default.

  ## Breaking Change 

  **TLS is now enabled by default.** 

  If you upgrade to this version and don't have TLS configured, you may see:
  rootCertPaths must not be empty

  **New users** must either:
  - Configure TLS with `rootCerts`, or
  - Explicitly set `tls.enabled: false`

  **Existing users** can add `tls.enabled: false` to restore previous behavior.                                                                                                            
   
  ## Changes                                                                                                                                                                               
                  
  - TLS now defaults to `enabled: true` instead of `false`
  - `Normalize()` sets `Enabled=true` when nil
  - TLS warning moved to config load time (single centralized warning instead of per-client)
  - Integration tests updated with explicit `tls.enabled: false` (testcontainers use `--insecure` flag)
  - Added `assert.False` message for Queries TLS assertion for consistency

  ## Migration

  **Enable TLS (recommended for production):**
  ```yaml
  tls:
    enabled: true
    rootCerts:
      - /path/to/ca.crt

  Disable TLS (not recommended for production):
  tls:
    enabled: false
```
 Fixes #108
  Test Plan

  - Unit tests pass
  - Integration tests pass

### DIFF
--- a/tools/fxconfig/internal/config/config.go
+++ b/tools/fxconfig/internal/config/config.go
@@ -60,7 +60,7 @@ type MSPConfig struct {
 //
 //nolint:revive,lll
 type TLSConfig struct {
-	Enabled            *bool    `mapstructure:"enabled" yaml:"enabled,omitempty" desc:"Enable/disable TLS" default:"false"`
+	Enabled            *bool    `mapstructure:"enabled" yaml:"enabled,omitempty" desc:"Enable/disable TLS" default:"true"`
 	ClientKeyPath      string   `mapstructure:"clientKey" yaml:"clientKey,omitempty" desc:"Path to TLS client private key"`
 	ClientCertPath     string   `mapstructure:"clientCert" yaml:"clientCert,omitempty" desc:"Path to TLS client certificate"`
 	RootCertPaths      []string `mapstructure:"rootCerts" yaml:"rootCerts,omitempty" desc:"Paths to TLS root certificates"`
@@ -68,10 +68,10 @@ type TLSConfig struct {
 }
 
 // Normalize ensures the TLS config has an explicit enabled flag.
-// Sets enabled to false if not specified.
+// Sets enabled to true if not specified.
 func (c *TLSConfig) Normalize() {
 	if c.Enabled == nil {
-		enabled := false
+		enabled := true
 		c.Enabled = &enabled
 	}
 }
--- a/tools/fxconfig/internal/config/load_test.go
+++ b/tools/fxconfig/internal/config/load_test.go
@@ -423,6 +423,8 @@ queries:
 
 notifications:
   address: localhost:7002
+  tls:
+    enabled: false
 `
 	err := os.WriteFile(configPath, []byte(configContent), 0o600)
 	require.NoError(t, err)
--- a/tools/fxconfig/internal/cli/v1/root.go
+++ b/tools/fxconfig/internal/cli/v1/root.go
@@ -11,10 +11,13 @@ package v1
 import (
 	"github.com/spf13/cobra"
 
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	"github.com/hyperledger/fabric-x/tools/fxconfig/internal/app"
 	"github.com/hyperledger/fabric-x/tools/fxconfig/internal/cli/v1/cliio"
 	"github.com/hyperledger/fabric-x/tools/fxconfig/internal/config"
 )
+
+var rootLogger = flogging.MustGetLogger("root")
 
 // NewRootCommand constructs and returns the root command for fxconfig.
 // It sets up configuration loading, flag registration, and all subcommands.
@@ -48,6 +51,12 @@ Configuration can be provided via:
 			cfg, err := config.Load(opts...)
 			if err != nil {
 				return err
+			}
+
+			// Warn if TLS is disabled for any service (secure-by-default)
+			if !cfg.Orderer.TLS.IsEnabled() || !cfg.Queries.TLS.IsEnabled() || !cfg.Notifications.TLS.IsEnabled() {
+				rootLogger.Warn("TLS is disabled for one or more services — " +
+					"connections will be unencrypted. This is insecure and not recommended for production.")
 			}
 
 			// set our config and printer in our context
--- a/tools/fxconfig/internal/config/config_test.go
+++ b/tools/fxconfig/internal/config/config_test.go
@@ -44,7 +44,7 @@ func TestTLSConfig_Normalize(t *testing.T) {
 		cfg      *TLSConfig
 		wantBool bool
 	}{
-		{"nil Enabled", &TLSConfig{}, false},
+		{"nil Enabled", &TLSConfig{}, true},
 		{"already false", &TLSConfig{Enabled: boolPtr(false)}, false},
 		{"already true", &TLSConfig{Enabled: boolPtr(true)}, true},
 	}

### CODE REVIEW

File: tools/fxconfig/internal/config/load_test.go [hunk 1]
```diff
@@ -435,7 +435,7 @@ notifications:
 	assert.True(t, cfg.Orderer.TLS.IsEnabled())
 	assert.Equal(t, []string{"/path/to/ca.pem"}, cfg.Orderer.TLS.RootCertPaths)
 	assert.False(t, cfg.Queries.TLS.IsEnabled())
-	assert.False(t, cfg.Notifications.TLS.IsEnabled())
+	assert.True(t, cfg.Notifications.TLS.IsEnabled())
```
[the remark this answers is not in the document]
mbrandenburger: We need to update the tests now to disable TLS explicitly. Also update the documentation accordingly.

mbrandenburger (changes requested):

mbrandenburger:
hi @MayankSharmaCSE it seems that the unit tests are still failing.

MayankSharmaCSE:
@mbrandenburger, fixed the issue .

tools/fxconfig/internal/config/load_test.go
```diff
@@ -435,7 +437,7 @@ notifications:
 	assert.True(t, cfg.Orderer.TLS.IsEnabled())
 	assert.Equal(t, []string{"/path/to/ca.pem"}, cfg.Orderer.TLS.RootCertPaths)
 	assert.False(t, cfg.Queries.TLS.IsEnabled())
-	assert.False(t, cfg.Notifications.TLS.IsEnabled())
+	assert.False(t, cfg.Notifications.TLS.IsEnabled(), "Notifications explicitly has enabled: false in config")
```
mbrandenburger: Should we add a similar comment to `assert.False(t, cfg.Queries.TLS.IsEnabled())` as well?

tools/fxconfig/internal/client/client.go
```diff
@@ -40,6 +40,8 @@ func createSecOpts(tlsConfig *config.TLSConfig) (*comm.SecureOptions, error) {
 
 	// let's see if we use TLS
 	if !tlsConfig.IsEnabled() {
+		logger.Warn("TLS is disabled — connections will be unencrypted and " +
+			"unauthenticated. This is insecure and not recommended for production.")
```
mbrandenburger: I am wondering if this is the right place to throw the logging as this would lead to multiple warning for each  client (orderer, queries, notifications). Without additional information which service has TLS disabled.

mbrandenburger: Maybe we should handle the warning somewhere else ... during config loading?

tools/fxconfig/docs/README.md
mbrandenburger, line 156: I think we need to document this better and explain that by default TLS is enabled and requires at least a `rootCert` otherwise the user will get an error. Or the user explicitly turns off TLS.

mbrandenburger (changes requested):
Thank you @MayankSharmaCSE for working on this PR. We are making good progress here. I have a few more comments below.

Moreover, we should better indicate that this is a breaking change. Please make sure that the PR description highlights this. New users need to configure TLS correctly or disable it explicitly. 

The commit message should therefore also contain `fix(fxconfig)!` per conventional commits.

tools/fxconfig/internal/config/load_test.go
MayankSharmaCSE: Yeh , I missed that during the refactor. i will add the assertion for cfg.Queries.TLS.IsEnabled() to ensure consistency across all three service configs in the test.

tools/fxconfig/internal/client/client.go
MayankSharmaCSE: Fair point. Logging in each client would produce redundant warnings without context. I agree a single warning during config load would be cleaner. However, to keep this PR focused, I did prefer to address this as a follow-up item rather than scope creep this fix.

tools/fxconfig/docs/README.md
MayankSharmaCSE: I will add clear documentation explaining:
- TLS is now enabled by default and requires TLS certificates.
- Users can set tls.enabled: false to disable (not recommended for production).
- The "rootCertPaths must not be empty" error indicates misconfiguration.                                                                                                                                                                                    This will be added to the README and/or inline comments in the config example file.

MayankSharmaCSE:
@mbrandenburger Thanks for the review you had! I have addressed all your comments:
                                                                                                                                                                                           
  1. Breaking change — Updated commit prefix to fix(fxconfig)!: and added a prominent Breaking Change section in the PR description warning users that TLS is now enabled by default.
  2. Queries assertion — Added the missing assert.False message for cfg.Queries.TLS.IsEnabled() for consistency with Orderer and Notifications.
  3. Warning logging location — Moved the TLS warning to ResolveTLS() in config.go so it logs once at config load time instead of multiple times per client. Happy to address this as a follow-up if you'd prefer a different approach.                                                                                                                                          
  4. Documentation — Added a Breaking Change notice in the README explaining the new default, the error users may see (rootCertPaths must not be empty), and how to migrate.

tools/fxconfig/internal/config/load_test.go
mbrandenburger, line 32: Let's add TLS enabled as assertions as it's our new default.

mbrandenburger: For instance:
```go
assert.True(t, cfg.Orderer.TLS.IsEnabled())
// ... (and Queries/Notifications)
```

tools/fxconfig/internal/config/config.go
```diff
@@ -40,6 +45,12 @@ func (c *Config) ResolveTLS() {
 
 	c.Notifications.TLS = c.Notifications.TLS.InheritFrom(&c.TLS)
 	c.Notifications.TLS.Normalize()
+
+	// Warn once if any service has TLS disabled
+	if !c.Orderer.TLS.IsEnabled() || !c.Queries.TLS.IsEnabled() || !c.Notifications.TLS.IsEnabled() {
+		configLogger.Warn("TLS is disabled for one or more services — " +
+			"connections will be unencrypted. This is insecure and not recommended for production.")
+	}
```
mbrandenburger: I have the feeling we should not do the logging here as the "application" should decide if this particular configuration should log a warning or not. Thus, checking the config and print a log could happen in https://github.com/hyperledger/fabric-x/blob/321cab8844c7e2019d12768b91894465c13a7e96/tools/fxconfig/internal/cli/v1/root.go#L55  WDYT?

mbrandenburger (changes requested):
Thank you @MayankSharmaCSE Great work! See my comments below.

mbrandenburger:
@pasquale95 Can you please have a look at this PR, which changes the "default TLS behavior" for `fxconfig` to always enabled - if not explicitly disabled.

tools/fxconfig/internal/config/load_test.go
MayankSharmaCSE: Makes sense. I will add assert.True for when TLS is enabled by default (not explicitly disabled).

tools/fxconfig/internal/config/config.go
MayankSharmaCSE: Agreed, the config package shouldn't dictate logging. I will move the TLS warning from `config.go`to `root.go` so the application controls when/how to log. Does that sound right to you?

mbrandenburger:
This PR was re-opened in https://github.com/hyperledger/fabric-x/pull/214

MayankSharmaCSE:
@mbrandenburger,yes,I have followed up on the changes in #214 because i messed up my git history with some wrong commands.

PR #141, opened 2026-04-17
fix: handle HTTP server startup errors in sample services

### LINKED ISSUE #113: HTTP server startup failures are silently ignored in sample services
In the sample services (owner, issuer, endorser), the HTTP server is started using `go s.ListenAndServe()`, and the returned error is not handled.

If the server fails to start (for example, due to the port already being in use), the goroutine exits silently while the main process continues running. This creates a misleading state where the service appears to be running but is not actually serving any requests.

## Steps to Reproduce

1. Start another process on the same port used by the sample service (e.g., `python3 -m http.server <port>`).
2. Run one of the Fabric-X sample services.
3. Observe that:
   - The process remains running
   - The API is not accessible
   - No startup error is reported

## Expected Behavior

The service should fail fast and report an error if the HTTP server cannot start (e.g., port binding failure).

## Actual Behavior

The service continues running without serving requests and without reporting any error.

## Root Cause

`ListenAndServe` is executed in a goroutine and its returned error is ignored, so failures during server startup are not propagated or logged.

## Impact
- Silent failure during startup
- Difficult debugging and diagnosis
- Misleading service health (process is alive but not functional)

## Suggested Fix

Capture and handle the error returned by `ListenAndServe`. If the server fails to start, the application should log the error and terminate or handle it explicitly.

aaradhychinche-alt:
## Summary

Fixes silent HTTP server startup failures in sample token services by properly handling errors returned from `ListenAndServe`.

## Problem

The owner, issuer, and endorser services start the HTTP server using `go s.ListenAndServe()` and ignore the returned error.

If the server fails to start (e.g., port already in use), the goroutine exits silently while the process continues running without serving requests.

## Root Cause

The error returned by `ListenAndServe` is not checked, causing startup failures to be ignored.

## Solution

* Wrap `ListenAndServe` in a goroutine that checks the returned error
* Ignore `http.ErrServerClosed` (expected during shutdown)
* For any other error:

  * log the failure
  * terminate the process (fail fast)

## Impact

* Eliminates silent startup failures
* Services now fail fast when server cannot start
* No change to normal startup or graceful shutdown behavior

## Testing

* Manually reproduced failure by running another process on the same port
* Verified that service now logs error and exits instead of silently running
* Ensured normal startup and shutdown behavior remain unchanged

## Fixes

Closes #113


aaradhychinche-alt:
@mbrandenburger  CI failure seems unrelated; likely due to flaky integration test
raising a separate issue for this !!

aaradhychinche-alt:
@mbrandenburger  all tests are passing please merge if it is ok ?

mbrandenburger:
@arner can you please have a look at this?

mbrandenburger:
Closing as #113 is closed.

PR #143, opened 2026-04-17
fix: stabilise duplicate namespace integration test under parallel execution

### LINKED ISSUE #142: Flaky integration test: TestSingleOrgScenarios/duplicate_namespace_creation_fails depends on unstable namespace list output
## Problem

The integration test `TestSingleOrgScenarios/duplicate_namespace_creation_fails` compares namespace list output using `require.Equal(expectedStdOut, stdOut)`.

However, all subtests in `TestSingleOrgScenarios` run with `t.Parallel()` while sharing the same environment. Other parallel subtests can create namespaces concurrently, causing the namespace list output to change between calls.

This makes the assertion unstable and leads to flaky failures.

## Observed Behavior

* Namespace list output differs between calls
* Additional namespaces appear in output due to parallel execution
* Test fails even though behavior is correct

## Additional Issue

The test name suggests duplicate creation should fail, but the test expects:

```go
require.NoError(t, err)
```

This contradicts the intended behavior.

## Suggested Fix

* Avoid strict string equality for namespace list output
* Assert presence of expected namespace instead of full output match
* OR disable parallel execution for this test
* OR isolate test environment per subtest

## Impact

* Causes CI instability
* Leads to unrelated PR failures
* Makes debugging harder for contributors

aaradhychinche-alt:
## Summary

Fixes a flaky integration test in `TestSingleOrgScenarios/duplicate_namespace_creation_fails` by replacing brittle full-output comparison with stable semantic validation.

## Problem

The test previously compared the full output of `namespace list` using strict equality:

* This output is not stable because subtests run with `t.Parallel()` and share the same environment
* Other subtests can create namespaces concurrently, causing the output to change between calls
* This leads to nondeterministic failures in CI

## Root Cause

The test relied on byte-for-byte equality of command output, which is not deterministic under parallel execution with shared state.

## Solution

* Removed full stdout equality comparison
* Introduced semantic validation:

  * Verify that the expected namespace exists
  * Ensure it appears exactly once (no duplicate creation)
* Use namespace name for comparison instead of full struct equality to avoid instability

## Impact

* Eliminates flaky behavior in integration tests
* Improves reliability of CI
* Preserves original test intent (no duplicate namespace creation)

## Context

This issue surfaced while working on #113, where unrelated CI failures exposed instability in this test.

## Testing

* Verified test logic locally
* Change is limited to test assertions (no production code affected)

## Notes

This change keeps the test parallel-safe without altering test structure or behavior beyond removing nondeterministic assumptions.

Closes #142

### DIFF
--- a/tools/fxconfig/integration/single_org_test.go
+++ b/tools/fxconfig/integration/single_org_test.go
@@ -161,21 +161,18 @@ func TestSingleOrgScenarios(t *testing.T) {
 			configArg, policyArg, endorseArg, submitArg)
 		require.NoError(t, err)
 
-		// expect out namespace to be installed
-		// we keep the stdOut
-		var expectedStdOut string
+		// expect our namespace to be installed
 		assert.EventuallyWithT(t, func(ct *assert.CollectT) {
 			stdOut, err = fxconfig(t, "namespace", "list", configArg)
 			require.NoError(ct, err)
 			nss, err = parseNamespaceList(stdOut)
 			require.NoError(ct, err)
 			require.Contains(ct, nss, expectedNs)
-			expectedStdOut = stdOut
 		}, eventuallyTimeout, eventuallyTick)
 
 		// now we try to run create with the namespace again,
 		// but we use a different policy, as namespace creation should fail,
-		// and we expect the previous stdOut when calling list
+		// and we expect the namespace list to still contain exactly one entry for it
 		_, err = fxconfig(t, "namespace", "create", expectedNs.Name,
 			configArg, policyArg, endorseArg, submitArg)
 		require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


mbrandenburger (changes requested):
Thank you @aaradhychinche-alt for creating this PR. Can you please rebase your branch and only include the commit that relevant here. Thanks

aaradhychinche-alt:
> Thank you @aaradhychinche-alt for creating this PR. Can you please rebase your branch and only include the commit that relevant here. Thanks

done!!

mbrandenburger:
@aaradhychinche-alt your branch still shows 7 commits :) not one.

aaradhychinche-alt:
> @aaradhychinche-alt your branch still shows 7 commits :) not one.

done sorry !!

mbrandenburger (approved):
Thank you @aaradhychinche-alt

PR #146, opened 2026-04-17
fix: replace panic calls with proper error returns in owner fsc service

Hashan25-Dev:
## What does this PR do?

This PR fixes the hard `panic()` calls in `samples/tokens/owner/service/fsc.go`. Right now, if `GetManagementService` or `NewOwner` fails (e.g. due to a misconfiguration), the entire node crashes instead of returning a proper error. Since `Balances()`, `Balance()`, and `GetTransactions()` already have `error` in their return signatures, we should just use that.

Fixes hyperledger/fabric-x-samples#8

## Changes

- `Balances()` — return `nil, fmt.Errorf(...)` instead of panicking when management service init fails
- `Balance()` — return `Amount{}, fmt.Errorf(...)` instead of panicking
- `GetTransactions()` — return proper errors for both the management service failure and the `NewOwner` nil check

## Why?

Panicking in exported API functions that already return errors is bad practice. A single bad config value or transient issue shouldn't bring down the whole process. Callers should be able to handle and log these errors gracefully.

## Testing

- Verified the code compiles cleanly with `go build`
- Only touched the 4 panic lines, no other changes

### DIFF
--- a/samples/tokens/owner/service/fsc.go
+++ b/samples/tokens/owner/service/fsc.go
@@ -59,7 +59,7 @@ var (
 func (f FabricSmartClient) Balances(ctx context.Context, wallet string) ([]Amount, error) {
 	mgmt, err := token.GetManagementService(f.node)
 	if err != nil {
-		panic("configuration error, check the logs")
+		return nil, fmt.Errorf("configuration error, check the logs: %w", err)
 	}
 
 	wal := mgmt.WalletManager().OwnerWallet(ctx, wallet)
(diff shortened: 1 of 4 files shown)


aaradhychinche-alt:
@Hashan25-Dev  please solve the conflicts !!

iliecirciumaru:
Hi @Hashan25-Dev 
Thanks a lot for your contributions! Could you please move your code from sample folder into https://github.com/hyperledger/fabric-x-samples and close this PR? We are migrating all sample applications there. I apologize for this inconvenience

mbrandenburger:
Thank you @Hashan25-Dev  Closing this PR as we moved the samples to https://github.com/hyperledger/fabric-x-samples

PR #152, opened 2026-04-17
fix: properly run fxconfig version through global Printer interface

### LINKED ISSUE #151: `fxconfig version` command bypasses the CLI Printer interface

Hashan25-Dev:
Closes #151 

## Description

This PR knocks out a bit of technical debt inside the `fxconfig version` command. Previously, the command completely bypassed the global `CLIContext.Printer` interface that the rest of the application uses and just hardcoded raw `cmd.Printf` calls to stream output. 

This cleans up the dangling `// TODO: use printer` comment in the codebase, and prevents the issue of the version command breaking format entirely if we were to route the printer into JSON or custom output destinations later!

## Changes made
- Updated the `NewVersionCommand` signature to take `CLIContext`, matching our other CLI tools.
- Stripped out all the raw cobra `cmd.Printf` lines and mapped everything directly to `cliCtx.Printer.Print()`.
- Refactored the internal `showLine` helper into a functionally pure `formatLine` method so it just builds strings instead of forcing print side-effects.
- Overhauled the testing setups in `version_test.go` to properly inject a mocked `CLIContext`. Our tests now properly read from the injected `cliio.Printer` buffer rather than hijacking Cobra's `.SetOut()` method.

## How to test
Check out this branch locally and verify the outputs are completely visually identical, just mapped correctly under the hood:
```bash
go build -o fxconfig ./tools/fxconfig
./fxconfig version

### DIFF
--- a/samples/tokens/owner/service/fsc.go
+++ b/samples/tokens/owner/service/fsc.go
@@ -59,7 +59,7 @@ var (
 func (f FabricSmartClient) Balances(ctx context.Context, wallet string) ([]Amount, error) {
 	mgmt, err := token.GetManagementService(f.node)
 	if err != nil {
-		panic("configuration error, check the logs")
+		return nil, fmt.Errorf("configuration error, check the logs: %w", err)
 	}
 
 	wal := mgmt.WalletManager().OwnerWallet(ctx, wallet)
@@ -90,7 +90,7 @@ func (f FabricSmartClient) Balances(ctx context.Context, wallet string) ([]Amoun
 func (f FabricSmartClient) Balance(ctx context.Context, wallet, code string) (Amount, error) {
 	mgmt, err := token.GetManagementService(f.node)
 	if err != nil {
-		panic("configuration error, check the logs")
+		return Amount{}, fmt.Errorf("configuration error, check the logs: %w", err)
 	}
 	wal := mgmt.WalletManager().OwnerWallet(ctx, wallet)
 	if wal == nil {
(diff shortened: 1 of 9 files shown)


FirePheonix:
again man, this was done in my PR: https://github.com/hyperledger/fabric-x/pull/82 .
and this has 9 file changes.
i think you've merged multiple PRs in one PR?

image

iliecirciumaru:
Hi @Hashan25-Dev 
Thanks a lot for your contributions! Could you please move your code from samples folder into https://github.com/hyperledger/fabric-x-samples and close this PR? We are migrating all sample applications there. I apologize for this inconvenience

mbrandenburger:
Thank you @Hashan25-Dev for creating this PR. #151 is already closed. Thanks again

PR #153, opened 2026-04-17
fxconfig: clean up provider-managed grpc clients at app shutdown

### LINKED ISSUE #116: fxconfig: prevent reuse of closed orderer and notification clients across submissions
Title: fxconfig: prevent reuse of closed orderer and notification clients across submissions

Problem

`fxconfig` uses `provider.Provider.Get()` to lazily initialize and cache service instances with `sync.Once`.

However, the submission path closes some of the returned clients after use:
- the orderer client is closed at the end of submission
- the notification client is also closed in the `--wait` flow

This creates an ownership mismatch:
- providers cache these clients as long-lived instances
- callers treat them as per-operation resources

As a result, repeated operations may reuse a cached client whose underlying connection has already been closed.

Code path

- `tools/fxconfig/internal/provider/provider.go`: `Provider.Get()` memoizes instances using `sync.Once`
- `tools/fxconfig/internal/app/submit.go`: submission closes the orderer client after use
- `tools/fxconfig/internal/app/submit.go`: `SubmitTransactionWithWait` also closes the notification client

Why it matters

This may cause repeated CLI operations in the same application instance to fail unpredictably and makes client lifecycle semantics unclear.

Suggested direction

Choose one ownership model and apply it consistently:

1. Providers should not cache closeable gRPC clients and should create fresh instances per use, or
2. Providers should own the lifecycle of cached clients, and consumers should not close provider-managed instances.

Acceptance criteria

- repeated submit operations work correctly
- repeated submit-with-wait operations work correctly
- client ownership semantics are clear in code
- regression tests cover repeated usage

Kumar-s75:
#### Type of change

- Bug fix

#### Description

This PR fixes grpc client lifecycle ownership in `fxconfig`.

Currently, `fxconfig` caches grpc clients through providers, but some app-layer operations close those clients after use. In particular, the submission and query paths treat provider-managed clients as per-operation resources, which creates an ownership mismatch between the provider layer and its consumers.

This change keeps the existing provider-based caching model intact and moves cleanup responsibility to application shutdown instead.

The main changes are:

- add `Close()` to the `app.Application` interface
- implement `AdminApp.Close()` to release provider-managed resources
- add `Provider.Close()` to close cached instances that implement `Close()`
- remove per-operation `Close()` calls from:
  - `tools/fxconfig/internal/app/submit.go`
  - `tools/fxconfig/internal/app/list.go`
- update `main.run()` so the CLI closes the application once command execution completes

This makes lifecycle ownership explicit:

- providers own cached grpc clients
- app operations use them without closing them
- the CLI triggers shutdown once and cleanup happens there

This follows the existing architecture more closely and avoids ambiguity around resource ownership.

Regression coverage was added for:

- repeated submit operations using provider-managed orderer clients
- repeated `--wait` operations using provider-managed orderer and notification clients
- repeated query operations using provider-managed query clients
- cleanup of cached clients during app shutdown
- provider cleanup behavior for initialized and uninitialized instances

Validation performed:

```bash
go test -v ./tools/fxconfig/internal/provider ./tools/fxconfig/internal/app ./tools/fxconfig/internal/cli/v1
```

Also checked:
```bash
git diff --check
```
A full go test ./tools/fxconfig/... run is currently blocked in my local environment by missing integration fixtures:

tools/fxconfig/integration/testdata/crypto/single-org.pb.bin
tools/fxconfig/integration/testdata/crypto/multi-org.pb.bin

Related issues
Closes #116

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 
 # python
 .venv
+
+.codex
--- a/tools/fxconfig/internal/app/app.go
+++ b/tools/fxconfig/internal/app/app.go
@@ -10,6 +10,8 @@ package app
 
 import (
 	"context"
+	"errors"
+	"sync"
 
 	"github.com/hyperledger/fabric-x-common/api/applicationpb"
 	"github.com/hyperledger/fabric-x-common/msp"
@@ -27,6 +29,7 @@ type Application interface {
 	SubmitTransaction(ctx context.Context, txID string, tx *applicationpb.Tx) error
 	SubmitTransactionWithWait(ctx context.Context, txID string, tx *applicationpb.Tx) (TxStatus, error)
 	MergeTransactions(ctx context.Context, txs []*applicationpb.Tx) (*applicationpb.Tx, error)
+	Close() error
 }
 
 // AdminApp implements Application interface with provider-based dependencies.
@@ -36,4 +39,30 @@ type AdminApp struct {
 	QueryProvider        *provider.Provider[adapters.QueryClient, *config.QueriesConfig]
 	OrdererProvider      *provider.Provider[adapters.OrdererClient, *config.OrdererConfig]
 	NotificationProvider *provider.Provider[adapters.NotificationClient, *config.NotificationsConfig]
+	closeOnce            sync.Once
+	closeErr             error
+}
+
+// Close releases provider-managed resources owned by the application.
+func (d *AdminApp) Close() error {
+	d.closeOnce.Do(func() {
+		var errs []error
+
+		if d.NotificationProvider != nil {
+			errs = append(errs, d.NotificationProvider.Close())
+		}
+		if d.OrdererProvider != nil {
+			errs = append(errs, d.OrdererProvider.Close())
+		}
+		if d.QueryProvider != nil {
+			errs = append(errs, d.QueryProvider.Close())
+		}
+		if d.MspProvider != nil {
+			errs = append(errs, d.MspProvider.Close())
+		}
+
+		d.closeErr = errors.Join(errs...)
+	})
+
+	return d.closeErr
 }
--- a/tools/fxconfig/internal/cli/v1/namespace_create_test.go
+++ b/tools/fxconfig/internal/cli/v1/namespace_create_test.go
@@ -141,3 +141,13 @@ func (t *testApp) SubmitTransactionWithWait(
 	args := t.Called(ctx, txID, tx)
 	return args.Int(0), args.Error(1)
 }
+
+func (t *testApp) Close() error {
+	for _, call := range t.ExpectedCalls {
+		if call.Method == "Close" {
+			args := t.Called()
+			return args.Error(0)
+		}
+	}
+	return nil
+}
--- a/tools/fxconfig/internal/provider/provider.go
+++ b/tools/fxconfig/internal/provider/provider.go
@@ -21,6 +21,8 @@ type Provider[T any, K Validatable] struct {
 	instance          T
 	err               error
 	cfg               K
+	initialized       bool
+	ready             bool
 	validationContext validation.Context
 }
 
@@ -42,18 +44,37 @@ func New[T any, K Validatable](
 // Subsequent calls return the cached instance. Thread-safe.
 func (p *Provider[T, K]) Get() (T, error) {
 	p.once.Do(func() {
+		p.initialized = true
 		if err := p.cfg.Validate(p.validationContext); err != nil {
 			p.err = err
 			return
 		}
+
 		p.instance, p.err = p.factory(p.cfg)
+		p.ready = p.err == nil
 	})
+
 	return p.instance, p.err
 }
 
 // Validate delegates to the configuration's Validate method.
 func (p *Provider[T, K]) Validate() error {
 	return p.cfg.Validate(p.validationContext)
+}
+
+// Close releases the cached instance if it was initialized successfully and
+// implements a Close() error method. Non-closeable instances are ignored.
+func (p *Provider[T, K]) Close() error {
+	if !p.initialized || !p.ready {
+		return nil
+	}
+
+	closer, ok := any(p.instance).(interface{ Close() error })
+	if !ok {
+		return nil
+	}
+
+	return closer.Close()
 }
 
 // Validatable defines the interface for configuration types that can be validated.
--- a/tools/fxconfig/main.go
+++ b/tools/fxconfig/main.go
@@ -8,6 +8,7 @@ package main
 
 import (
 	"context"
+	"errors"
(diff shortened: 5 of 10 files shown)


mbrandenburger:
Than you @Kumar-s75 for creating this PR!

> A full go test ./tools/fxconfig/... run is currently blocked in my local environment by missing integration fixtures:
> 
> tools/fxconfig/integration/testdata/crypto/single-org.pb.bin tools/fxconfig/integration/testdata/crypto/multi-org.pb.bin

Did you try to run `make generate`? This should produce the missing test artifacts.

Please also have a look at the linter errors (you can use `make lint`).

.codex
mbrandenburger, line 1: This file should not be committed

tools/fxconfig/internal/app/app.go
```diff
@@ -37,3 +39,23 @@ type AdminApp struct {
...
+		errs = append(errs, d.OrdererProvider.Close())
+	}
+	if d.QueryProvider != nil {
+		errs = append(errs, d.QueryProvider.Close())
+	}
+	if d.MspProvider != nil {
+		errs = append(errs, d.MspProvider.Close())
+	}
+
+	return errors.Join(errs...)
```
mbrandenburger: Probably we should guard this Close with a sync.Once. WDYT?

mbrandenburger (changes requested):
Thank you @Kumar-s75 for working on this PR. Can you please make the linter happy. More comments below.

Kumar-s75:
@mbrandenburger Thank you! for reviewing this PR,i have made all the required changes and i agree as i updated `AdminApp.Close()` to be idempotent using `sync.Once` and added a regression test to verify repeated `Close()` calls only release managed resources once.

tools/fxconfig/main.go
```diff
@@ -74,5 +75,10 @@ func run() error {
 		},
 	)
 
-	return cmd.ExecuteContext(ctx)
+	err := cmd.ExecuteContext(ctx)
+	if cliCtx.App != nil {
+		err = errors.Join(err, cliCtx.App.Close())
+	}
```
mbrandenburger, line 81: What about moving this to a `PersistentPostRun` function in `cli/v1/root.go`? WDYT?

.gitignore
```diff
@@ -7,4 +7,6 @@
 .ansible
 
 # python
-.venv
\ No newline at end of file
+.venv
+
+.codex
```
mbrandenburger, line 12: This change should not be in this PR. I suggest that you add `.codex` file on your local global `.gitignore`.

tools/fxconfig/internal/provider/provider.go
```diff
@@ -21,6 +21,8 @@ type Provider[T any, K Validatable] struct {
 	instance          T
 	err               error
 	cfg               K
+	initialized       bool
+	ready             bool
```
mbrandenburger, line 25: I suggest we replace ready with `atomic.Bool` to avoid race conditions. Then there is no need for `initialized`.

tools/fxconfig/internal/provider/provider.go
```diff
@@ -42,12 +44,16 @@ func New[T any, K Validatable](
...
 func (p *Provider[T, K]) Get() (T, error) {
 	p.once.Do(func() {
+		p.initialized = true
 		if err := p.cfg.Validate(p.validationContext); err != nil {
 			p.err = err
 			return
 		}
+
 		p.instance, p.err = p.factory(p.cfg)
+		p.ready = p.err == nil
```
mbrandenburger, line 54: With our ready becoming `atomic.Bool` we can do the following.

```suggestion
		if p.err == nil {
				p.ready.Store(true)		
		}
```

tools/fxconfig/internal/provider/provider.go
```diff
@@ -56,6 +62,21 @@ func (p *Provider[T, K]) Validate() error {
 	return p.cfg.Validate(p.validationContext)
 }
 
+// Close releases the cached instance if it was initialized successfully and
+// implements a Close() error method. Non-closeable instances are ignored.
+func (p *Provider[T, K]) Close() error {
+	if !p.initialized || !p.ready {
+		return nil
+	}
```
mbrandenburger, line 70:
```suggestion
	if !p.ready.Load() {
		return nil
	}
```

tools/fxconfig/internal/app/app.go
```diff
@@ -36,4 +39,30 @@ type AdminApp struct {
...
+		}
+		if d.OrdererProvider != nil {
+			errs = append(errs, d.OrdererProvider.Close())
+		}
+		if d.QueryProvider != nil {
+			errs = append(errs, d.QueryProvider.Close())
+		}
+		if d.MspProvider != nil {
+			errs = append(errs, d.MspProvider.Close())
+		}
```
mbrandenburger, line 62: This becomes a bit messy ... the msp instance does not have a `close()`. However, our generic `Provider` now has a close and class the instance close if it is closeable. We can drop the `mspProvider.Close()` here. WDYT?

tools/fxconfig/internal/cli/v1/namespace_create_test.go
```diff
@@ -141,3 +141,13 @@ func (t *testApp) SubmitTransactionWithWait(
...
 }
+
+func (t *testApp) Close() error {
+	for _, call := range t.ExpectedCalls {
+		if call.Method == "Close" {
+			args := t.Called()
+			return args.Error(0)
+		}
+	}
+	return nil
```
mbrandenburger, line 152: I am wondering if this is the correct way to do ... would it be something like that ....

```go
func (t *testApp) Close() error {
      args := t.Called()
      return args.Error(0)
  }
```
and use https://pkg.go.dev/github.com/stretchr/testify/mock#Call.Maybe on the mock setup? WDYT?

mbrandenburger (changes requested):
Thank you @Kumar-s75 for working on this. 
While I still think the approach we discussed in issue #116—introducing the instance lifecycle to the provider—makes sense in principle, this PR highlights the amount of additional code complexity it brings. As always, the real challenges will surface when we start testing, particularly around avoiding race conditions.

Given that, I’d like to revisit the approach. The core problem we’re trying to solve is that application code using a client instance via the lazy provider should not call close, since the provider effectively owns the instance. If we simply remove the corresponding defer close calls, that issue is resolved.

The tradeoff is that we lose graceful shutdown of the gRPC clients. I’d question whether that is actually a problem in our case. Since we’re dealing with a short-lived CLI that exits after executing a few commands, the gRPC stack should clean up its resources on termination anyway.

So this becomes a question of balancing code complexity against best practices—I'm not convinced the added lifecycle management is worth it here. What do you think?

PR #169, opened 2026-04-18
fix(fxconfig): cleaning up stale subscriber on canceled/timeout Subscribe paths

### LINKED ISSUE #109: fxconfig: Subscribe leaks stale subscriber on canceled context

rootp1:
## Problem

`NotificationClient.Subscribe()` appended a subscriber before validating context state. If the context was canceled before queueing the upstream request, the method returned with an error but left stale subscriber state in memory for the same txID.

## Root cause

In `Subscribe()` the map mutation happened too early:

1. Create receiver channel
2. Append subscriber to `subscribers[txID]`
3. Later fail on canceled/timed-out context before successful queue send
4. Stale subscriber remained

## Implemented fix

Updated `Subscribe()` to keep single-upstream-subscription behavior and clean up stale state on failed enqueue:

- Keep subscriber registration under lock
- If this is not the first subscriber for a txID, return immediately (reuse existing upstream subscription)
- For first subscriber path, if context is canceled before enqueue, delete `subscribers[txID]`
- If context is canceled while waiting to enqueue, delete `subscribers[txID]`

This guarantees the map is rolled back when request enqueue does not happen.

### DIFF
--- a/tools/fxconfig/internal/client/notifications.go
+++ b/tools/fxconfig/internal/client/notifications.go
@@ -91,7 +91,6 @@ func (n *NotificationClient) Subscribe(ctx context.Context, txID string) (chan i
 	}
 
 	receiverCh := make(chan int, 1)
-
 	n.subscribersMu.Lock()
 	defer n.subscribersMu.Unlock()
 
@@ -114,6 +113,7 @@ func (n *NotificationClient) Subscribe(ctx context.Context, txID string) (chan i
 	// check if our ctx is still open
 	select {
 	case <-ctx.Done():
+		delete(n.subscribers, txID)
 		return nil, ctx.Err()
 	default:
 	}
@@ -121,6 +121,7 @@ func (n *NotificationClient) Subscribe(ctx context.Context, txID string) (chan i
 	// try to push to request queue
 	select {
 	case <-ctx.Done():
+		delete(n.subscribers, txID)
 		return nil, ctx.Err()
 	case n.requestQueue <- req:
 	}
--- a/tools/fxconfig/internal/client/notifications.go
+++ b/tools/fxconfig/internal/client/notifications.go
@@ -92,15 +92,16 @@ func (n *NotificationClient) Subscribe(ctx context.Context, txID string) (chan i
 
 	receiverCh := make(chan int, 1)
 	n.subscribersMu.Lock()
-	defer n.subscribersMu.Unlock()
 
 	subscribers := n.subscribers[txID]
 	n.subscribers[txID] = append(subscribers, receiverCh)
 
 	if len(subscribers) > 0 {
 		// we already have an active subscription for this txID
+		n.subscribersMu.Unlock()
 		return receiverCh, nil
 	}
+	n.subscribersMu.Unlock()
 
 	// setup request
 	req := &committerpb.NotificationRequest{
@@ -113,7 +114,7 @@ func (n *NotificationClient) Subscribe(ctx context.Context, txID string) (chan i
 	// check if our ctx is still open
 	select {
 	case <-ctx.Done():
-		delete(n.subscribers, txID)
+		n.removeSubscriber(txID, receiverCh)
 		return nil, ctx.Err()
 	default:
 	}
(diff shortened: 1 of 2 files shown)


rootp1:
@mbrandenburger, Please have a look, whenever you have a chance to. Thankyou

tools/fxconfig/internal/client/notifications.go
```diff
@@ -114,13 +113,15 @@ func (n *NotificationClient) Subscribe(ctx context.Context, txID string) (chan i
...
 	case <-ctx.Done():
+		delete(n.subscribers, txID)
 		return nil, ctx.Err()
 	default:
 	}
 
 	// try to push to request queue
 	select {
 	case <-ctx.Done():
+		delete(n.subscribers, txID)
```
mbrandenburger: this deletion comes with the problem that we just remove all subscribers, if there are any and not just the one we are adding here.

mbrandenburger (changes requested):
Thank you @rootp1 for creating this PR. I think this is good start. 

Please have a look at https://github.com/hyperledger/fabric-x/pull/123. Maybe we can cherry pick the changes to `notifications.go` and add you tests. WDYT?

tools/fxconfig/internal/client/notifications.go
rootp1: Thanks for calling this out @mbrandenburger, I’ve addressed it in the latest update.

In notifications.go, I replaced the broad delete(n.subscribers, txID) rollback with targeted cleanup that removes only the subscriber channel created by the current Subscribe call. 

If other subscribers already exist for the same txID, they are preserved.

I also added a regression test to verify that canceling one blocked subscribe path does not remove existing subscribers for the same txID.

Please take another look when you have a moment.

rootp1:
> Thank you @rootp1 for creating this PR. I think this is good start.
> 
> Please have a look at #123. Maybe we can cherry pick the changes to `notifications.go` and add you tests. WDYT?

As, discussed I am up with the changes suggested by you @mbrandenburger. Please have a look whenever you have a chance to

tools/fxconfig/internal/client/notifications.go
rootp1: also, my regression test verifies confirms that when one blocked Subscribe call is canceled, only that call’s subscriber is removed, and any other existing subscriber for the same txID remains registered.

cynox-66:
Hey @rootp1 
This looks much better now — the targeted cleanup makes sense 👍

One thing I was thinking about while reading this: `removeSubscriber` and the dispatcher in `listen()` both end up touching the same map. The locking + guard makes it safe, but it might be worth a quick comment there just to make that interaction obvious for anyone revisiting this later.

Also nice catch on the regression test — that’s a subtle edge case.

rootp1:
> Hey @rootp1 This looks much better now — the targeted cleanup makes sense 👍
> 
> One thing I was thinking about while reading this: `removeSubscriber` and the dispatcher in `listen()` both end up touching the same map. The locking + guard makes it safe, but it might be worth a quick comment there just to make that interaction obvious for anyone revisiting this later.
> 
> Also nice catch on the regression test — that’s a subtle edge case.

Thankyou.

tools/fxconfig/internal/client/notifications.go
```diff
@@ -114,20 +114,51 @@ func (n *NotificationClient) Subscribe(ctx context.Context, txID string) (chan i
...
 	case <-ctx.Done():
+		n.removeSubscriber(txID, receiverCh)
 		return nil, ctx.Err()
 	case n.requestQueue <- req:
 	}
 
 	return receiverCh, nil
 }
 
+func (n *NotificationClient) removeSubscriber(txID string, receiverCh chan int) {
```
mbrandenburger: I liked the the small inline rollback function @cynox-66 used in his PR. In particular, because it's only used within `Subscribe()`. Do you see any advantage to promote `removeSubscriber` to a method of `NotificationClient`?

```go
rollback := func() {
....    
}
```

tools/fxconfig/internal/client/notifications.go
```diff
@@ -91,17 +91,17 @@ func (n *NotificationClient) Subscribe(ctx context.Context, txID string) (chan i
...
 
 	subscribers := n.subscribers[txID]
 	n.subscribers[txID] = append(subscribers, receiverCh)
 
 	if len(subscribers) > 0 {
 		// we already have an active subscription for this txID
+		n.subscribersMu.Unlock()
 		return receiverCh, nil
 	}
+	n.subscribersMu.Unlock()
```
mbrandenburger: replacing `defer Unlock()` with two explicit `Unlock()` seem to work but we could mess this code up later easily. Do you think something like the following would work?

```go
  isFirst := func() bool {
      n.subscribersMu.Lock()
      defer n.subscribersMu.Unlock()
      existing := n.subscribers[txID]
      // update the subscribers 
      n.subscribers[txID] = append(existing, receiverCh)
      return len(existing) == 0
  }()
  if !isFirst {
      return receiverCh, nil
  }
```

mbrandenburger (changes requested):
Thank you @rootp1 for working on this PR! Good progress. I left a few comments below.

tools/fxconfig/internal/client/notifications.go
rootp1: I agree with keeping this local to Subscribe(). 
I refactored it into an inline rollback closure since it’s only used on the subscribe path and doesn’t add reusable behavior as a client-wide method. This also keeps subscription mutation and rollback logic co-located.

one possible advantage of a NotificationClient method is reuse if the same cleanup is needed from multiple call sites, but right now keeping it local in Subscribe() is clearer and safer because it captures only function-local context and avoids widening the public/internal API surface.

tools/fxconfig/internal/client/notifications.go
rootp1: Thankyou for the suggestion @mbrandenburger, I switched to exactly that pattern via an inline isFirst function (Lock + defer Unlock) and early return when it’s not the first subscriber. 

That removes the double explicit unlock path and makes the locking logic safer to maintain.

rootp1:
@mbrandenburger, please have a look whenever you have a chance to.

mbrandenburger:
@cynox-66 could you also have another look please?

mbrandenburger (approved):

cynox-66:
Hey @mbrandenburger  — took another look after the latest changes.

The `isFirst` pattern + inline rollback makes the flow much cleaner now, and the targeted cleanup fixes the original issue without affecting existing subscribers.
The regression test also covers the tricky cancellation case well.

Looks good from my side 👍

mbrandenburger:
Thank you @rootp1 @cynox-66 great work!

rootp1:
> Thank you @rootp1 @cynox-66 great work!

Thankyou @mbrandenburger, for the appreciation

cynox-66:
Thank you @mbrandenburger .

PR #170, opened 2026-04-18
fxconfig: fail fast when notification stream startup fails

### LINKED ISSUE #154: fxconfig: fail fast when notification stream startup fails
Title: fxconfig: fail fast when notification stream startup fails

Problem

`NewNotificationClient()` starts the notification listener in a background goroutine and immediately returns success.

If the underlying notification stream fails to open, the error is only logged/printed from the background goroutine and is not returned to the caller. This means callers can receive a client whose stream is already unusable.

Location

- `tools/fxconfig/internal/client/notifications.go`

Why it matters

This can lead to delayed or misleading failures in flows that depend on notifications, especially `SubmitTransactionWithWait()`. Instead of surfacing a clear initialization failure, the caller may proceed with a client that cannot receive events.

Suggested direction

Possible approaches:
1. make `NewNotificationClient()` synchronously verify initial stream setup and return an error on failure, or
2. track listener health and make `Subscribe()` fail immediately if the stream is not healthy

Acceptance criteria

- notification stream startup failures are returned clearly to callers
- notification-dependent flows fail fast with actionable errors
- regression tests cover failed stream initialization

Kumar-s75:
#### Type of change

- Bug fix
- Test update

#### Description

This PR fixes notification client initialization in `fxconfig` so that notification stream startup failures are surfaced immediately.

Currently, `NewNotificationClient()` starts the listener in a background goroutine and returns success before the underlying notification stream is confirmed to be usable. If the stream fails to open, the caller can still receive a client object even though the stream is already broken.

This change makes notification client creation fail fast instead.

The main changes are:

- update `NewNotificationClient()` to wait for initial stream establishment before returning success
- add an internal startup handshake so stream initialization errors are returned directly to the caller
- preserve `streamErr` handling so later stream failures still cause `Subscribe()` to fail fast
- add regression tests for failed notification stream startup and startup error propagation

This improves reliability for notification-dependent flows such as `SubmitTransactionWithWait()` by ensuring callers do not proceed with a client whose stream was never established.

Regression coverage was added for:

- notification client creation failing fast when stream startup fails
- startup errors being stored and observable on the client
- existing subscribe / wait behavior remaining intact

Validation performed:

```bash
go test -v ./tools/fxconfig/internal/client
go test -v ./tools/fxconfig/internal/app
```

Also checked:
```bash
git diff --check
```

Closes #154

### DIFF
--- a/tools/fxconfig/internal/app/app.go
+++ b/tools/fxconfig/internal/app/app.go
@@ -10,6 +10,7 @@ package app
 
 import (
 	"context"
+	"errors"
 
 	"github.com/hyperledger/fabric-x-common/api/applicationpb"
 	"github.com/hyperledger/fabric-x-common/msp"
@@ -27,6 +28,7 @@ type Application interface {
 	SubmitTransaction(ctx context.Context, txID string, tx *applicationpb.Tx) error
 	SubmitTransactionWithWait(ctx context.Context, txID string, tx *applicationpb.Tx) (TxStatus, error)
 	MergeTransactions(ctx context.Context, txs []*applicationpb.Tx) (*applicationpb.Tx, error)
+	Close() error
 }
 
 // AdminApp implements Application interface with provider-based dependencies.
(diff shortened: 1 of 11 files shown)


mbrandenburger:
Thank you @Kumar-s75 for creating this PR. Closing this as we closed #154

PR #77, opened 2026-04-02
feat(fxconfig): add configurable bccsp for msp signer

Sukuna0007Abhi:
#### Type of change

- New feature
- Improvement (improvement to code, performance, etc)
- Test update
- Documentation update

#### Description

This PR adds configurable BCCSP support to fxconfig for MSP signer setup.

Instead of hardcoding software BCCSP defaults in the signer path, the MSP config now supports a bccsp section and uses it to instantiate the crypto provider correctly.
The implementation also keeps sensible defaults (SW, SHA2, 256, keystore under MSP path) when values are not provided.

Included in this PR:

- BCCSP config structs under msp config
- Conversion logic to Fabric FactoryOpts
- Signer update to use config-driven BCCSP instantiation
- Tests for defaults, overrides, and config loading
- Docs update with config examples

#### Additional details (Optional)

- Added new config models under `msp.bccsp`:
  - `default`
  - `sw.security`
  - `sw.hash`
  - `sw.fileKeyStore.keyStorePath`
- Added conversion helper from fxconfig config model to Fabric `factory.FactoryOpts`
- Updated signer setup to:
  - pass BCCSP opts to `GetLocalMspConfig`
  - instantiate provider via `factory.GetBCCSPFromOpts`
- Added tests for:
  - default BCCSP mapping
  - override mapping
  - config file loading of BCCSP settings
  - signer initialization with explicit BCCSP config
- Updated docs with example `msp.bccsp` configuration

Testing performed:

- `go test ./tools/fxconfig/internal/config -run "TestMSPConfigToFactoryOpts_Defaults|TestMSPConfigToFactoryOpts_Overrides|TestLoad_WithConfigFile"`
- `go test ./tools/fxconfig/internal/msp -run "TestGetSignerIdentityFromMSP"`
- `go build ./tools/fxconfig`

#### Related issues

Related to #60

PR #177, opened 2026-04-20
feat(fxconfig): add PKCS#11 / HSM BCCSP support

tinystone007:
#### Type of change

- New feature (non-breaking)

#### Description

fxconfig currently hard-codes a file-based software BCCSP that reads keys from `<ConfigPath>/keystore`. Operators who keep MSP private keys in an HSM or an external KMS have no way to configure fxconfig to sign transactions without first exporting keys to disk, which defeats the purpose of the HSM.

This PR adds optional PKCS#11 support, mirroring the configuration shape already used by Fabric / Fabric-X peers and orderers.

## Changes

- Extend `MSPConfig` with a `BCCSPConfig` section containing a `PKCS11Config` (library path, label, pin, hash family, security level, and the usual `softwareVerify` / `immutable` flags).
- When `MSPConfig.BCCSP.PKCS11.Library` is non-empty, `setupMSP` builds a PKCS#11 BCCSP via `fabric-lib-go/bccsp/pkcs11`, backed by a no-op key store (keys live inside the HSM, so file-based key management is unnecessary).
- Otherwise behaviour is unchanged: fxconfig falls back to the existing file-based software BCCSP.

No existing callers are affected — the new fields are zero-value by default and trigger the original code path. An operator opts in by setting `BCCSP.PKCS11.Library` in their fxconfig config.

## Example config

```yaml
msp:
  localMspID: Org1MSP
  configPath: /msp
  bccsp:
    pkcs11:
      library: /usr/local/lib/libsofthsm2.so
      label: TokenLabel
      pin: ${HSM_PIN}
```

## Files changed

- `tools/fxconfig/internal/config/config.go` — add `BCCSPConfig` / `PKCS11Config` structs.
- `tools/fxconfig/internal/msp/signer.go` — build BCCSP based on config; keep file-based path as default.

## Test plan

- [x] Built fxconfig with and without PKCS#11 config; default path unchanged.
- [x] Verified fxconfig signs namespace transactions against an HSM-backed MSP (SoftHSM2) end-to-end in a downstream integration environment.
- [x] CI.

### DIFF
--- a/tools/fxconfig/internal/config/config.go
+++ b/tools/fxconfig/internal/config/config.go
@@ -51,8 +51,29 @@ type LoggingConfig struct {
 // MSPConfig contains MSP (Membership Service Provider) identity configuration.
 // It specifies which organization identity to use for signing transactions.
 type MSPConfig struct {
-	LocalMspID string `mapstructure:"localMspID" yaml:"localMspID,omitempty" desc:"MSP ID of the organization"`
-	ConfigPath string `mapstructure:"configPath" yaml:"configPath,omitempty" desc:"Path to MSP configuration directory"`
+	LocalMspID string      `mapstructure:"localMspID" yaml:"localMspID,omitempty" desc:"MSP ID of the organization"`
+	ConfigPath string      `mapstructure:"configPath" yaml:"configPath,omitempty" desc:"Path to MSP configuration directory"`
+	BCCSP      BCCSPConfig `mapstructure:"bccsp" yaml:"bccsp,omitempty" desc:"BCCSP (crypto provider) configuration; leave empty for default file-based keystore"`
+}
+
+// BCCSPConfig configures the BCCSP (Blockchain Crypto Service Provider).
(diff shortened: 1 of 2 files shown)


mbrandenburger:
@tinystone007 Thank for creating this PR. It seems that there is already another PR addressing your suggestion. Can you please have a look at https://github.com/hyperledger/fabric-x/pull/77 If you think something is missing there let's work on improving that PR. Thanks

tinystone007:
Closing in favour of a stacked PR against #77 per @mbrandenburger's earlier suggestion — reworked on top of @Sukuna0007Abhi's `BCCSPConfig` / `ToFactoryOpts` scaffolding so the PKCS#11 support lands as an extension of #77 rather than a competing PR.

Stacked PR: https://github.com/Sukuna0007Abhi/fabric-x/pull/1

If #77 merges and the maintainers would rather have this as a follow-up against `main`, happy to reopen against `main` here instead.

PR #179, opened 2026-04-21
fix(fxconfig): unblock subscribers and fail fast when notification stream terminates

### LINKED ISSUE #109: fxconfig: Subscribe leaks stale subscriber on canceled context

### LINKED ISSUE #164: Notification subscription state corruption on subscribe timeout/cancel leads to missed events
## Description
A failed (timed-out or canceled) subscription attempt can leave a stale entry in the internal `subscribers` map, causing subsequent subscriptions for the same `txID` to skip sending a new upstream request and never receive events.

This results in a broken notification flow where `WaitForEvent` may block or timeout indefinitely, even though the system is otherwise functioning correctly.

## Root Cause
In `Subscribe`, the subscriber channel is added to `n.subscribers` before ensuring that the request is successfully enqueued.

If the enqueue operation fails due to timeout or context cancellation, the function returns early without rolling back the previously inserted state.

**Relevant locations:**
- Insert into map: `tools/fxconfig/internal/client/notifications.go:98-100`
- Early return on timeout/cancel: `tools/fxconfig/internal/client/notifications.go:123-125`
- Duplicate-subscription fast path: `tools/fxconfig/internal/client/notifications.go:101-104`

Because of this:
- A stale subscriber remains registered
- Future calls detect an existing subscriber and skip enqueueing a new request
- No actual notification request is sent upstream

## Steps to Reproduce
1. Create a `NotificationClient` with:
   - a very short `WaitingTimeout`
   - no active consumer for `requestQueue` (to force enqueue failure)
2. Call:
```go
Subscribe(ctx, "tx1")
```
→ returns context deadline exceeded
3. Call again:
```go
Subscribe(ctx, "tx1")
```
### Observe:
- The second call returns a channel.
- No new request is sent to `requestQueue`.
- Call `WaitForEvent` on the returned channel:
  - It blocks or times out.
  - No event is ever delivered.

## Expected Behavior:
- Failed subscription attempts should not mutate internal state.
- A subsequent `Subscribe` call for the same `txID` should perform a fresh upstream request.

## Actual Behavior:
- Failed first attempt leaves a stale subscriber entry.
- Subsequent subscriptions reuse invalid state and skip request enqueue.
- Event delivery for that `txID` is permanently broken.

## Impact:
- Breaks `--wait` semantics for transaction finality,
- Causes indefinite waits or timeouts for specific `txID`s,
- Difficult to debug because failure depends on timing and backpressure,
- Can occur under transient conditions (slow consumer, short timeout).
## Severity
Critical

## Suggested Fix:
Ensure that state mutation is rolled back if enqueue fails.

### Minimal approach:
1. Insert subscriber under lock,
2. Release lock before enqueue,
3. If enqueue fails:
   - Reacquire lock,
   - Delete the inserted subscriber channel,
   - Delete the `txID` entry if empty.

### Edge Cases:
- Concurrent subscribers for the same `txID` must still deduplicate correctly.
- Rollback must not remove valid subscribers added concurrently.
- Ensure no race conditions between enqueue and rollback paths.

sridhar-panigrahi:
I was reading through the notification client code and noticed that when the listener goroutine exits — say the committer restarts or the gRPC stream drops — callers end up in a pretty bad state.

**What's broken today:**

The cleanup in `listen()` calls `clear(n.subscribers)` but never closes the subscriber channels. So if you're in the middle of a `WaitForEvent` call when the stream dies, you just sit there until your own `WaitingTimeout` expires — which can be a long time. Meanwhile, any new `Subscribe` call after the listener exits blocks forever on the unbuffered `requestQueue` because nobody's reading from it anymore. The client becomes a zombie that silently hangs everything.

There's also a leftover `fmt.Printf` on the error path even though the package already has a `flogging` logger declared in `client.go`.

**What this PR does:**

- Adds a `done chan struct{}` to `NotificationClient`, closed by `listen()` before it clears the map.
- During cleanup, closes every subscriber channel so callers in `wait()` get `ErrNotificationStreamClosed` immediately instead of hanging.
- `Subscribe()` checks `done` under the map lock (to serialize with cleanup) and also on the `requestQueue` select, returning the same sentinel error. Holding the lock for the `done` check is important — it prevents a race where a new subscriber appends to the map after cleanup already ran.
- `wait()` uses a two-value channel receive to distinguish a closed channel from an actual status delivery.
- Replaces `fmt.Printf` with `logger.Errorf` using the existing package logger.
- Fixes `orderer.go` to use `%s` when formatting the rejection status so you get `FORBIDDEN` instead of `0xb`.

**Tests:**

Two new tests alongside the existing ones:
- `TestNotificationClient_ListenExit_UnblocksWaitForEvent` — simulates listener exit and verifies `WaitForEvent` returns `ErrNotificationStreamClosed` right away, not after the 10-second timeout.
- `TestNotificationClient_SubscribeAfterListenExit_ReturnsSentinel` — verifies `Subscribe` returns the sentinel immediately after listener death rather than blocking.

All existing unit tests pass. Integration tests require Docker so they weren't run locally.

Fixes #109
Fixes #164
Partial progress on #154

Signed-off-by: Shridhar Panigrahi <sridharpanigrahi2006@gmail.com>

### DIFF
--- a/tools/fxconfig/internal/client/notifications.go
+++ b/tools/fxconfig/internal/client/notifications.go
@@ -19,6 +19,12 @@ import (
 	"github.com/hyperledger/fabric-x/tools/fxconfig/internal/config"
 )
 
+// ErrNotificationStreamClosed is returned when the notification stream has
+// terminated, either because the gRPC stream failed or the client was closed.
+// Callers receive this from Subscribe or WaitForEvent instead of blocking until
+// their own context deadline expires.
+var ErrNotificationStreamClosed = errors.New("notification stream closed")
+
 // NotificationClient provides a gRPC client for receiving transaction status notifications.
 // It manages bidirectional streaming with the committer notification service and multiplexes
 // notifications to multiple subscribers per transaction ID.
@@ -32,6 +38,10 @@ type NotificationClient struct {
 
 	subscribers   map[string][]chan int
 	subscribersMu sync.RWMutex
+
+	// done is closed when listen() returns. Subscribers use it to fail fast
+	// instead of blocking on requestQueue or their own context deadline.
+	done chan struct{}
 }
 
 // NewNotificationClient creates a notification client with the provided configuration.
@@ -54,12 +64,12 @@ func NewNotificationClient(cfg config.NotificationsConfig) (*NotificationClient,
 		requestQueue:  make(chan *committerpb.NotificationRequest),
 		responseQueue: make(chan *committerpb.NotificationResponse),
 		subscribers:   make(map[string][]chan int),
+		done:          make(chan struct{}),
 	}
 
 	go func() {
 		if err := nc.listen(ctx); err != nil && !errors.Is(err, context.Canceled) {
-			fmt.Printf("error: Notification listener stream terminated unexpectedly: %s\n", err)
-			// logger.Errorf("Notification listener stream terminated unexpectedly for %s: %s", key, err)
+			logger.Errorf("Notification listener stream terminated unexpectedly: %s", err)
 		}
 	}()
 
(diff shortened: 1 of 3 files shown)


sridhar-panigrahi:
Closing — I didn't check issue assignments and existing PRs before opening this. #109 is already being worked on by @rootp1 (PR #123) and #154 by @Kumar-s75 (PR #170). Apologies for the noise.

aaradhychinche-alt:
@sridhar-panigrahi  please do no raise a pr on the issues i am working on and first taking reviews on the issues then raising a pr

PR #97, opened 2026-04-09
feat(fxmigrate): scaffold Exporter CLI for Fabric to Fabric-X state migration

sridhar-panigrahi:
This adds `tools/fxmigrate` — the Exporter CLI that handles the export side of the Fabric → Fabric-X migration pipeline from RFC hyperledger/fabric-x-rfcs#5 and epic #21.

The tool reads a standard Fabric peer snapshot directory, filters out everything that doesn't belong in Fabric-X (PDC hashes, system chaincodes, implicit org collections), converts Fabric version pairs to Fabric-X scalar integers, and writes a genesis-data file that the committer's `--init-from-snapshot` bootstrap mode can ingest.

## Structure

Follows the same layout as `fxconfig` — cobra CLI, internal packages, unit tests:

```
tools/fxmigrate/
  main.go
  internal/
    cli/v1/
      root.go       cobra root: export / verify / version subcommands
      export.go     5-phase export flow
      verify.go     placeholder — post-migration verifier is a follow-up PR
      version.go
    snapshot/
      manifest.go   reads _snapshot_signable_metadata.json, recomputes SHA-256 checksums
      reader.go     streaming binary parser for public_state.data
      filter.go     namespace and key filtering rules
    genesis/
      writer.go     length-delimited binary output with JSON header
```

## What the export command does

```
fxmigrate export \
  --snapshot ./peer/snapshots/completed/mychannel/100 \
  --channel  mychannel \
  --namespace token \
  --output   genesis.bin
```

Five phases:
1. Read `_snapshot_signable_metadata.json` and verify SHA-256 checksums for all snapshot files
2. Scan `public_state.data` to discover which namespaces are present
3. Stream all key-value pairs through the filter (drops PDC, system chaincodes, CouchDB internals)
4. Write the genesis-data file — length-delimited binary with a JSON header containing namespace, source channel, block height, entry count, and export timestamp
5. Print SHA-256 of the output file

## Filtering

Per RFC §3.1, excluded from migration:
- System chaincodes: `lscc`, `_lifecycle`, `cscc`, `qscc`, `escc`, `vscc`
- PDC hash namespaces (`$$h$$` pattern) and PDC private state (`$$p$$` pattern)
- Implicit org collections (any namespace containing `$$`)
- CouchDB internal metadata keys (`\x00collection` prefix)

## Version conversion

Fabric stores versions as `(blockNum, txNum)` pairs. Converted to Fabric-X scalar BIGINT: `(blockNum << 32) | txNum`

## Tests

[...]

Issue #21, opened 2025-10-31
Define Data Migration Path from Hyperledger Fabric Snapshot to Fabric-X


cendhu:
**Problem Statement & Motivation**

To provide a viable upgrade path for existing Hyperledger Fabric users, we need to define a process for migrating ledger data from a Fabric network to Fabric-X.

The official peer ledger snapshot feature in Hyperledger Fabric (as documented [here](https://hyperledger-fabric.readthedocs.io/en/release-2.5/peer_ledger_snapshot.html)) seems like the best starting point, as it provides a verifiable copy of the world state at a specific block height.

However, a direct 1-to-1 import is not possible due to key architectural differences:

1. **No Channels**: Fabric-X does not have the concept of channels. Fabric data is namespaced by channel. We must define how data from one or more channels maps into the single state model of Fabric-X.

2. **No Private Data**: Fabric-X does not have private data collections (PDCs). The Fabric snapshot contains hashes of private data, which are irrelevant to Fabric-X and must be excluded.

This issue is to design the export/import process and define the target data set for this migration.

**Analysis of Fabric Snapshot Data**

Based on the documentation, a Fabric snapshot directory contains several key data components:

 - **Public State (World State)**: The complete key-value state from the state database (LevelDB or CouchDB). This is the primary data to migrate.

 - **Private Data Hashes**: Hashes of data stored in PDCs. This must be ignored.

 - **Transactions IDs**: A list of all processed transaction IDs. This is used to prevent replay attacks. This should be considered for migration.

 - **Collection Config History**: The history of PDC configurations. This must be ignored.

 - Metadata Files (_snapshot_signable_metadata.json): Contains channel name, block height, and hashes of the snapshot files for verification.

**Proposed Tasks & Acceptance Criteria**

This effort needs to be broken down into two main parts: (1) defining the data export and mapping, and (2) defining the Fabric-X import mechanism.

**Part 1: Data Export & Mapping Definition**

- [ ] Define the Canonical Data Set: Specify exactly which data from the Fabric snapshot will be part of the migration.

- [ ] Resolve the Channel-to-Global State Mapping: This is the most critical design decision. A strategy must be defined for handling data from multiple Fabric channels.

  - Option 1 (Prefixing): The migration tool prefixes all keys with their channel of origin (e.g., mychannel_namespace_key123). This preserves all data in a single Fabric-X instance but may require chaincode modification.

[...]

PR #180, opened 2026-04-21
feat(fxmigrate): implement verify command and genesis file reader

sridhar-panigrahi:
Follow-up to #97 — implements the placeholder `verify` subcommand that was scaffolded there.

## What this adds

**`genesis.Reader`** (`internal/genesis/reader.go`)
The reader is the inverse of the existing `genesis.Writer`. It parses the same length-delimited binary format — `uint32`-prefixed JSON header followed by key/value/version triples — and exposes two functions: `ReadHeader` for a quick header-only check, and `ReadAll` for loading the full set of entries.

**`fxmigrate verify --genesis <file> --snapshot <dir>`**
Runs four integrity checks in sequence:

1. Parses the genesis-data file and prints the header metadata (namespace, source channel, block height, sha256).
2. Counts entries actually present in the file and compares against `header.EntryCount` — catches a truncated or partially-written output file before it ever reaches the committer.
3. Reads the source snapshot manifest and checks that the block height recorded in the genesis header matches.
4. Re-runs the same PDC/system-chaincode filtering over the snapshot and compares the filtered entry count against the genesis count — confirms nothing was silently dropped or duplicated during export.

This gives operators a way to sanity-check the artifact produced by `export` before handing it to the committer bootstrap, without needing the committer to be running at all.

## Tests

6 round-trip tests in `genesis/reader_test.go`: header round-trip, full entry round-trip, empty file, missing file (both `ReadHeader` and `ReadAll`), binary values.

## Related

- Epic: #21
- Export CLI: #97

Signed-off-by: Shridhar Panigrahi <sridharpanigrahi2006@gmail.com>

### DIFF
--- a/tools/fxconfig/internal/client/notifications_test.go
+++ b/tools/fxconfig/internal/client/notifications_test.go
@@ -184,3 +184,193 @@ func TestNotificationClient_Close_NilFunc(t *testing.T) {
 	nc := &NotificationClient{}
 	require.NoError(t, nc.Close())
 }
+
+// Concurrent subscriber tests
+
+func TestNotificationClient_ConcurrentSubscribers_SameTxID_BothReceive(t *testing.T) {
+	t.Parallel()
+
+	nc := newTestNotificationClient(2 * time.Second)
+
+	// First subscriber — drains the request queue so Subscribe doesn't block
+	go func() { <-nc.requestQueue }()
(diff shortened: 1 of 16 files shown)


sridhar-panigrahi:
@cendhu  , @mbrandenburger , please let me know your thoughts on this !

PR #196, opened 2026-04-23
feat(fxconfig): parse namespace policies into human-readable strings

### LINKED ISSUE #195: fxconfig namespace list displays policies as raw hex instead of human-readable strings
in ```list.go```
Image

The fxconfig namespace list command currently outputs namespace policies as raw hexadecimal bytes, making it impossible to understand the endorsement policy without manual decoding. A parsePolicy function was stubbed out in internal/app/list.go but left entirely commented out with a // TODO: some pretty print would be beautiful note and *left for future implementation*

Expected behavior:

we need to implement ```parsePolicy``` with proper error handling, add ```signaturePolicyToString``` to recursively decode the ```NOutOf``` policy tree, wire ```PolicyStr``` into ```NamespaceQueryResult``` 
Then we'd have to update ```namespace_list.go``` to print ```p.PolicyStr``` instead of the ```%x p.Policy```.

FirePheonix:
#### Type of change
- New feature

#### Description
 This PR implements that ```TODO: parsePolicy now unmarshals the NamespacePolicy proto``` and returns either a DSL expression like ```OR('Org1MSP.member', 'Org2MSP.member')``` for MSP policies (by recursively walking the ```NOutOf tree``` via signaturePolicyToString) or ```Threshold(ECDSA, <base64-key>)``` for threshold policies, falling back to hex only if decoding fails. 

```namespace_list.go``` is updated to print ```p.PolicyStr``` and ```list_test.go``` covers OR, AND, OutOf(n).

 Threshold, invalid bytes, and end-to-end policy string population.

#### Related issues
- Fixes: #195

### DIFF
--- a/tools/fxconfig/internal/app/list.go
+++ b/tools/fxconfig/internal/app/list.go
@@ -8,12 +8,20 @@ package app
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
+	"strings"
+
+	cb "github.com/hyperledger/fabric-protos-go-apiv2/common"
+	msppb "github.com/hyperledger/fabric-protos-go-apiv2/msp"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/hyperledger/fabric-x-common/api/applicationpb"
 )
 
 // ListNamespaces queries the committer service for installed namespaces.
-// It connects to the query service, retrieves all namespace policies, and formats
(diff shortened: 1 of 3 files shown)


FirePheonix:
@mbrandenburger please review this, lemme know if this relevant to solve??

PR #203, opened 2026-04-26
fix(fxconfig): print TxID when using --wait on namespace create/update

### LINKED ISSUE #197: fxconfig: TxID is lost when using --wait on namespace create/update
## Summary

When running `fxconfig namespace create` or `fxconfig namespace update` with
`--endorse --submit --wait`, the Transaction ID (TxID) is silently discarded.
The user only sees a status message with no way to reference the transaction later.

## What happens

After `--wait`, the CLI prints:
Transaction status: COMMITTED

But the TxID is never shown. Without it, users have no way to track or audit
which transaction was submitted.

## What should happen
Transaction ID: <txid>
Transaction status: COMMITTED

## Root cause

In `tools/fxconfig/internal/app/deploy.go`, the `DeployNamespace` function
has the TxID available in `out.TxID` but explicitly discards it before returning:

```go
if input.Wait {
    status, err := d.SubmitTransactionWithWait(ctx, out.TxID, out.Tx)
    if err != nil {
        return nil, UnknownStatus, err
    }
    return nil, status, nil  // out.TxID is lost here
}
```

The same happens for the submit-without-wait path:

```go
if err := d.SubmitTransaction(ctx, out.TxID, out.Tx); err != nil {
    return nil, UnknownStatus, err
}
return nil, UnknownStatus, nil  // TxID lost here too
```

Because `res` comes back as `nil`, the CLI handlers in `namespace_create.go`
and `namespace_update.go` only print the status and skip the TxID entirely.

Compare this to running without `--wait` — the full transaction JSON including
TxID is printed. So the TxID is only visible when you don't actually submit.

## Proposed fix

In `deploy.go`, return the TxID even when submitting:

```go
return &DeployNamespaceOutput{TxID: out.TxID}, status, nil
```

Then update the CLI handlers in `namespace_create.go` and `namespace_update.go`
to print the TxID alongside the status:

```go
if res != nil && res.TxID != "" {
    ctx.Printer.Print(fmt.Sprintf("Transaction ID: %s", res.TxID))
}
ctx.Printer.Print(fmt.Sprintf("Transaction status: %s", committerpb.Status_name[int32(status)]))
```

## Files affected

- `tools/fxconfig/internal/app/deploy.go`
- `tools/fxconfig/internal/cli/v1/namespace_create.go`
- `tools/fxconfig/internal/cli/v1/namespace_update.go`

## Additional notes

I verified this by reading the source directly. Happy to submit a PR with the fix.

sublimemonk:
#### Type of change

- Bug fix

#### Description

When you run `fxconfig namespace create` or `fxconfig namespace update` with
`--endorse --submit --wait`, the transaction goes through successfully but the
TxID is never shown to the user. It was being discarded in `deploy.go` before
it could reach the CLI output.

This means after waiting for a transaction to commit, you only see:

    Transaction status: COMMITTED

With no way to know which transaction that was or reference it later.

The fix is straightforward — instead of returning nil from DeployNamespace after
the wait, we now return the TxID so both CLI handlers can print it alongside
the status:

    Transaction ID: <txid>
    Transaction status: COMMITTED

Three files changed:
- `deploy.go` — stop discarding the TxID on the wait and submit paths
- `namespace_create.go` — print TxID before the status message
- `namespace_update.go` — same fix

#### Related issues

Fixes #197

### DIFF
--- a/tools/fxconfig/internal/app/deploy.go
+++ b/tools/fxconfig/internal/app/deploy.go
@@ -86,11 +86,11 @@ func (d *AdminApp) DeployNamespace(
 		if err != nil {
 			return nil, UnknownStatus, err
 		}
-		return nil, status, nil
+		return &DeployNamespaceOutput{TxID: out.TxID}, status, nil
 	}
 	if err := d.SubmitTransaction(ctx, out.TxID, out.Tx); err != nil {
 		return nil, UnknownStatus, err
 	}
 
-	return nil, UnknownStatus, nil
+	return &DeployNamespaceOutput{TxID: out.TxID}, UnknownStatus, nil
 }
--- a/tools/fxconfig/internal/app/deploy_test.go
+++ b/tools/fxconfig/internal/app/deploy_test.go
@@ -166,7 +166,8 @@ func TestDeployNamespace_EndorseAndSubmit(t *testing.T) {
 
 	out, status, err := a.DeployNamespace(t.Context(), input)
 	require.NoError(t, err)
-	require.Nil(t, out)
+	require.NotNil(t, out)
+	require.NotEmpty(t, out.TxID)
 	require.Equal(t, UnknownStatus, status)
 }
 
@@ -205,7 +206,8 @@ func TestDeployNamespace_EndorseAndSubmitWithWait(t *testing.T) {
 
 	out, status, err := a.DeployNamespace(t.Context(), input)
 	require.NoError(t, err)
-	require.Nil(t, out)
+	require.NotNil(t, out)
+	require.NotEmpty(t, out.TxID)
 	require.Equal(t, expectedStatus, status)
 }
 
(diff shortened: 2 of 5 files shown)


et --soft HEAD~4
```diff
@@ -0,0 +1,65 @@
+[33m5fff2b6[m[33m ([m[1;36mHEAD[m[33m -> [m[1;32mpatch-1[m[33m, [m[1;31morigin/patch-1[m[33m)[m Update README.mdFix README: update Go version and correct sample app link  - Update Go requirement to 1.26+ - Remove outdated golangci-lint note - Replace external sample token link with local path (./samples/tokens)
```
iliecirciumaru: Hi,
I am a little bit confused, what is this file about ?

sublimemonk: Hi, That file was accidentally created during a git reset command while working on this branch. It's been removed now — the PR only has the actual fix. Sorry for the confusion!

PR #207, opened 2026-04-27
Update fabric-x-common

liran-funaro:
#### Type of change

#### Description

- Update fabric-x-common dependency


pasquale95 (approved):
LGTM!

PR #214, opened 2026-04-29
fix(fxconfig)!: enable TLS by default to prevent plaintext fallback vulnerability

### LINKED ISSUE #108: gRPC clients fall back to plaintext when TLS is disabled - no certificate validation

MayankSharmaCSE:
## Summary

  Enable TLS by default across all service configurations (orderer, queries, notifications). Previously TLS was disabled by default.

  ## Breaking Change 

  **TLS is now enabled by default.** 

  If you upgrade to this version and don't have TLS configured, you may see:
  rootCertPaths must not be empty

  **New users** must either:
  - Configure TLS with `rootCerts`, or
  - Explicitly set `tls.enabled: false`

  **Existing users** can add `tls.enabled: false` to restore previous behavior.                                                                                                            
   
  ## Changes                                                                                                                                                                               
                  
  - TLS now defaults to `enabled: true` instead of `false`
  - `Normalize()` sets `Enabled=true` when nil
  - TLS warning moved to config load time (single centralized warning instead of per-client)
  - Integration tests updated with explicit `tls.enabled: false` (testcontainers use `--insecure` flag)
  - Added `assert.False` message for Queries TLS assertion for consistency

  ## Migration

  **Enable TLS (recommended for production):**
  ```yaml
  tls:
    enabled: true
    rootCerts:
      - /path/to/ca.crt

  Disable TLS (not recommended for production):
  tls:
    enabled: false
```
 Fixes #108
  Test Plan

  - Unit tests pass
  - Integration tests pass

### DIFF
--- a/tools/fxconfig/docs/README.md
+++ b/tools/fxconfig/docs/README.md
@@ -141,6 +141,14 @@ notifications:
     enabled: false
 ```
 
+> **TLS is enabled by default (secure-by-default).** Each service (`orderer`, `queries`, `notifications`) must therefore either:
+> - provide at least `tls.rootCerts` (server-side TLS) — and additionally `tls.clientCert` / `tls.clientKey` for mutual TLS; or
+> - explicitly opt out with `tls.enabled: false`.
+>
+> A service with `tls.enabled: true` but no `rootCerts` will fail to load with `rootCertPaths must not be empty`.
+>
+> **Breaking change (v0.4.0+):** prior versions defaulted to `tls.enabled: false`. Configs upgraded from older versions that omit the `tls` section will now attempt TLS — set `tls.enabled: false` per service to keep the previous plaintext behavior.
+
 ### TLS Configuration
 
 - **No TLS**: `enabled: false` or all TLS fields empty
--- a/tools/fxconfig/integration/helpers_test.go
+++ b/tools/fxconfig/integration/helpers_test.go
@@ -140,15 +140,21 @@ orderer:
   address: ` + endpoints["orderer"] + `
   channel: ` + channelID + `
   connectionTimeout: 30s
+  tls:
+    enabled: false
 
 queries:
   address: ` + endpoints["query"] + `
   connectionTimeout: 20s
+  tls:
+    enabled: false
 
 notifications:
   address: ` + endpoints["sidecar"] + `
   connectionTimeout: 15s
   waitingTimeout: 15s
+  tls:
+    enabled: false
 `
 	err := os.WriteFile(configPath, []byte(configContent), 0o600)
 	require.NoError(tb, err)
--- a/tools/fxconfig/internal/cli/v1/root.go
+++ b/tools/fxconfig/internal/cli/v1/root.go
@@ -9,12 +9,15 @@ SPDX-License-Identifier: Apache-2.0
 package v1
 
 import (
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	"github.com/spf13/cobra"
 
 	"github.com/hyperledger/fabric-x/tools/fxconfig/internal/app"
 	"github.com/hyperledger/fabric-x/tools/fxconfig/internal/cli/v1/cliio"
 	"github.com/hyperledger/fabric-x/tools/fxconfig/internal/config"
 )
+
+var rootLogger = flogging.MustGetLogger("root")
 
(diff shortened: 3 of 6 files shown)


pasquale95:
@MayankSharmaCSE @mbrandenburger Is there a specific reason to default to TLS enabled? Isn't it counterintuitive? I think about `grpcurl` or `curl` which by default run without TLS verification.
Also as Copilot indicated now basically the user will have a fail message by just running `fxconfig` because he has to define rootCerts as a mandatory field.

MayankSharmaCSE:
@pasquale95 , i have few thoughts:                                                                                                                                                                          
                                            
  1. The decision to enable TLS by default follows security best practices. Tools like `grpcurl/curl` are general-purpose utilities that can work with or without TLS, but `fxconfig` manages Fabric-X infrastructure where security is critical. Making secure-by-default reduces the risk of misconfigurations in production.                                                                                   
  2. Users will see `rootCertPaths must not be empty` if they don't configure TLS, which clearly signals they need to act. The fix is simple: add `tls.enabled: false` to restore the previous behavior, or configure proper TLS certificates.                                                                                                                          
  3. Migration path - Existing users can simply add `tls.enabled: false` in their config to opt-out if they don't want TLS. The error message makes it obvious what needs to be fixed.

mbrandenburger:
> @MayankSharmaCSE @mbrandenburger Is there a specific reason to default to TLS enabled? Isn't it counterintuitive? I think about `grpcurl` or `curl` which by default run without TLS verification. Also as Copilot indicated now basically the user will have a fail message by just running `fxconfig` because he has to define rootCerts as a mandatory field.

@pasquale95 valid point. I strongly believe the TLS-enabled default is the right thing to move forward - however, it should be as easy as possible. What if we use the system root ca pool by default. If ca "custom" ca cert is needed; the corresponding error message will tell the user that ca server cert cannot be validated; hinting that custom certs can be set via `rootCertPaths`. If insecure operation mode is desired the user should just set `tls.enabled = false`.

WDYT?

pasquale95:
@mbrandenburger I believe this is a good approach. If TLS is enabled by default, then at least we try to use the system root CA pool if rootCerts is undefined.

PR #225, opened 2026-05-01
fix(fxconfig/validation): treat ".." only as a path component, not as…

abhayrajjais01:
#### Type of change

- Bug fix
- Test update

#### Description

`OSFileChecker` and `OSDirectoryChecker` rejected any cleaned path containing the substring `".."`, which blocked valid paths whose **file or directory names** include two dots (for example names like `foo..backup` or `config..yaml`), not actual traversal

**Change:** Replace the substring check with a rule that only treats leading `..` segments (whole path `".."` or prefix `".." + filepath.Separator`) as traversal, consistent with paths left by `filepath.Clean`

**Tests:** Extended `tools/fxconfig/internal/validation/validator_test.go` with cases for benign `..`-in-name paths, regressions (`..`, `../…`), and internal `..` segments that `Clean` collapses

**Suggested checks:**  
`go test -race ./tools/fxconfig/internal/validation/...`  
(or `go test -race ./tools/fxconfig/...` including integration tests if you prefer a full sweep)

### DIFF
--- a/tools/fxconfig/internal/validation/validator.go
+++ b/tools/fxconfig/internal/validation/validator.go
@@ -16,6 +16,20 @@ import (
 	"github.com/hyperledger/fabric-x-common/common/policydsl"
 )
 
+// hasTraversalComponent reports whether the cleaned path attempts to escape
+// its base via a ".." component. After filepath.Clean, the only way ".." can
+// remain is as a leading component (either the entire path or a "../" prefix);
+// any internal "..foo" or "foo..bar" segments are legitimate filename text
+// and must not be treated as traversal. This is intentionally stricter than
+// strings.Contains(clean, ".."), which produced false positives for any
+// filename whose name contained two consecutive dots.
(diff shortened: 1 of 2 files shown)


abhayrajjais01:
What was wrong?
`fxconfig` has checks that validate file paths and directory paths before doing things like loading certs or config.

Those checks tried to block path traversal (paths that sneak “upward” using `..`, like going outside the intended folder).

They did it in a clumsy way: “if the path string contains two dots (`..`), reject it.”


Why is that bad?
Lots of real, normal names accidentally contain `.` `.` together, without meaning “go up one folder”:

A backup folder named `myproject..backup`
A file named `config..yaml`
So the checker would say “path traversal not allowed” even when nothing dangerous was happening. Valid setups would be wrongly rejected.


What does the PR fix?
It narrows the rule:
Reject only real “go up” patterns (`..` at the start of the path, like `../secret`),
not every path that merely has `.` `.` somewhere in a name.

Issue #278, opened 2026-07-30
Tools: admin cli executable


tock-ibm:
Add under /tools the `main.go` of the admin cli (reconfig cli).
Add a target in the makefile to build it.

The code itself will be in fabric-x-common (for functionality common to orderer and committer) and in fabric-x-orderer (to reuse the Deliver/Broadcast client).

See: https://github.com/hyperledger/fabric-x-orderer/issues/1067

tock-ibm:
Assigned to: @MayRosenbaum

PR #282, opened 2026-08-18
fxadmin update readme

MayRosenbaum:
#### Type of change
update decode command description
update the compute-update command to include channel id
update the send command to require an output path for the prepared tx


tock-ibm (approved):

tock-ibm:
#278

PR #288, opened 2026-08-31
Added release-bins flow to release binaries in `*.tar.gz` archive on release tag

pasquale95:
#### Type of change

- New feature
- Documentation update

#### Description

Ship prebuilt `fabric-x-tools` CLI binaries (`configtxgen`, `configtxlator`, `cryptogen`, `fxconfig`) as GitHub Release assets whenever a `v*` tag is pushed, mirroring the existing `fabric-x-tools` image release but for users who don't want to run a container.

None of the four tools depend on cgo, so all platforms cross-compile from a single `ubuntu-latest` runner (no QEMU, no macOS runner needed):

- `linux/amd64`, `linux/arm64`, `linux/s390x` — same as the `fabric-x-tools` image
- `darwin/amd64`, `darwin/arm64` — for macOS users, since the image only ever runs on Linux

Changes:
- `Makefile`: new `release-bins` target that cross-compiles the tools via `go build` (the existing `bin/%` rule uses `go install`, which can't cross-compile) and stamps `Version`/`CommitSHA` via `METADATA_VAR` — previously unused, so released binaries now report their real version instead of `latest`.
- `scripts/create-binary-package.sh`: builds one `<os>-<arch>` target and packages it into `fabric-x-tools-<target>-<version>.tar.gz` (bin/ + LICENSE).
- `.github/workflows/release-binaries.yml`: new workflow, independent of `build-image.yml`. Validates the tag, builds the 5-platform matrix, generates `checksums.txt`, and publishes everything to the tag's GitHub Release via `softprops/action-gh-release@v2` (same action used by the `fabric-x-ansible-collection` release workflow). Also supports manual re-runs via `workflow_dispatch`.

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -1,0 +1,113 @@
+#
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+name: Release Binaries
+
+on:
+  push:
+    tags:
+      - "v*"
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag to release (e.g. v1.0.0). Re-runs an existing release."
+        required: true
+
+permissions:
+  contents: read
+
+jobs:
+  prepare:
+    name: Validate Tag
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.extract.outputs.tag }}
+      version: ${{ steps.extract.outputs.version }}
+    steps:
+      - name: Extract and validate version from tag
+        id: extract
+        env:
+          RAW_TAG: ${{ inputs.tag || github.ref_name }}
+        run: |
+          if [[ "${RAW_TAG}" =~ ^v([0-9]+\.[0-9]+\.[0-9]+.*)$ ]]; then
+            echo "tag=${RAW_TAG}" >> "$GITHUB_OUTPUT"
+            echo "version=${BASH_REMATCH[1]}" >> "$GITHUB_OUTPUT"
+            echo "✅ Tag validated: ${RAW_TAG}"
+          else
+            echo "❌ Invalid tag format: ${RAW_TAG}. Expected format: v<semver>"
+            exit 1
+          fi
+
+  build:
+    name: Build (${{ matrix.os }}/${{ matrix.arch }})
+    needs: prepare
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: linux
+            arch: amd64
+          - os: linux
+            arch: arm64
+          - os: linux
+            arch: s390x
+          - os: darwin
+            arch: amd64
+          - os: darwin
+            arch: arm64
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          ref: ${{ needs.prepare.outputs.tag }}
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - name: Build and package binaries
+        env:
+          TARGET: ${{ matrix.os }}-${{ matrix.arch }}
+          RELEASE: ${{ needs.prepare.outputs.tag }}
+          REVISION: ${{ github.sha }}
+        run: ./scripts/create-binary-package.sh
+
+      - name: Upload build artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: fabric-x-tools-${{ matrix.os }}-${{ matrix.arch }}
+          path: release/*/fabric-x-tools-*.tar.gz
+          if-no-files-found: error
+
+  release:
+    name: Publish GitHub Release
+    needs:
+      - prepare
+      - build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Download all build artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: dist
+          merge-multiple: true
+
+      - name: Generate checksums
+        working-directory: dist
+        run: sha256sum *.tar.gz > checksums.txt
+
+      - name: Publish release assets
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ needs.prepare.outputs.tag }}
+          files: |
+            dist/*.tar.gz
+            dist/checksums.txt
+          fail_on_unmatched_files: true
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -81,7 +81,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: fabric-x-tools-${{ matrix.os }}-${{ matrix.arch }}
-          path: release/*/fabric-x-tools-*.tar.gz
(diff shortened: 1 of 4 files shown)


.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - name: Build and package binaries
+        env:
+          TARGET: ${{ matrix.os }}-${{ matrix.arch }}
+          RELEASE: ${{ needs.prepare.outputs.tag }}
+          REVISION: ${{ github.sha }}
```
mbrandenburger, line 77: I am wondering if `github.sha` is correct here as it might not match the tag version we are trying to build here. If you just drop it here, your script will default to `git rev-parse HEAD`, which is in any case the correct checkout out version. Can we double check this somehow?

Makefile
```diff
@@ -63,9 +63,25 @@ $(BUILD_DIR)/%:
...
+RELEASE_DIR ?= release
+GOOS   ?= $(shell go env GOOS)
+GOARCH ?= $(shell go env GOARCH)
+RELEASE_BIN_DIR = $(RELEASE_DIR)/$(GOOS)-$(GOARCH)/bin
+
+.PHONY: release-bins
+release-bins: $(TOOLS_EXES:%=$(RELEASE_BIN_DIR)/%) ## Cross-compiles all tools for $(GOOS)/$(GOARCH) into $(RELEASE_DIR)
+
+$(RELEASE_DIR)/%: GO_LDFLAGS = $(METADATA_VAR:%=-X $(PKGNAME)/common/metadata.%)
+$(RELEASE_DIR)/%:
```
mbrandenburger, line 75: Do we need to add a `FORCE` prerequisite to ensure that we correctly rebuilding?!

Makefile
```diff
@@ -63,9 +63,25 @@ $(BUILD_DIR)/%:
...
+	@echo "Building $@"
+	@mkdir -p $(@D)
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) $(go_cmd) build -trimpath \
```
mbrandenburger, line 78: Can you check that these go flags are consistent with the docker-based builds, in particular the use of cgo?

Makefile
```diff
@@ -63,9 +63,25 @@ $(BUILD_DIR)/%:
 	@GOBIN=$(abspath $(@D)) go install -tags "$(GO_TAGS)" -ldflags "$(GO_LDFLAGS)" -buildvcs=false $(pkgmap.$(@F))
 	@touch $@
 
+RELEASE_DIR ?= release
+GOOS   ?= $(shell go env GOOS)
+GOARCH ?= $(shell go env GOARCH)
+RELEASE_BIN_DIR = $(RELEASE_DIR)/$(GOOS)-$(GOARCH)/bin
+
+.PHONY: release-bins
+release-bins: $(TOOLS_EXES:%=$(RELEASE_BIN_DIR)/%) ## Cross-compiles all tools for $(GOOS)/$(GOARCH) into $(RELEASE_DIR)
```
mbrandenburger, line 72: I am wondering what we should do with idemixgen as this is part of the docker-based build. It's kind of inconsistent.

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,113 @@
...
+          merge-multiple: true
+
+      - name: Generate checksums
+        working-directory: dist
+        run: sha256sum *.tar.gz > checksums.txt
+
+      - name: Publish release assets
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ needs.prepare.outputs.tag }}
```
mbrandenburger, line 109: depending on the tag, we may want to set `prerelease: true`, for instance, on a tag like `v1.2.0-rc1`. Something like `prerelease: ${{ contains(needs.prepare.outputs.tag, '-') }}` could help.

.github/workflows/release-binaries.yml
```diff
@@ -0,0 +1,113 @@
...
+permissions:
+  contents: read
+
+jobs:
+  prepare:
+    name: Validate Tag
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.extract.outputs.tag }}
+      version: ${{ steps.extract.outputs.version }}
```
mbrandenburger, line 28: do we use the `version` output?

scripts/create-binary-package.sh
```diff
@@ -0,0 +1,55 @@
...
+GOARCH="${BASH_REMATCH[2]}"
+
+# Trim the leading 'v' from the tag, matching build-image.yml's version extraction.
+VERSION="${RELEASE#v}"
+
+# Cross-compile via the Makefile, stamping version/commit into the binaries.
+RELEASE_DIR="release/${TARGET}"
+echo "Building ${TARGET} binaries for ${RELEASE} (${REVISION})..."
+make release-bins GOOS="${GOOS}" GOARCH="${GOARCH}" RELEASE_DIR=release \
+  METADATA_VAR="Version=${VERSION} CommitSHA=${REVISION}"
```
mbrandenburger, line 46: it seems that the script assumes we run it from the root folder. If invoked from the `script` folder it won't be happy.  Probably ok - maybe some more bash magic needed to make this more robust?

Makefile
```diff
@@ -63,9 +63,25 @@ $(BUILD_DIR)/%:
 	@GOBIN=$(abspath $(@D)) go install -tags "$(GO_TAGS)" -ldflags "$(GO_LDFLAGS)" -buildvcs=false $(pkgmap.$(@F))
 	@touch $@
 
+RELEASE_DIR ?= release
+GOOS   ?= $(shell go env GOOS)
+GOARCH ?= $(shell go env GOARCH)
```
mbrandenburger, line 68:
```suggestion
# Resolved on first use and then cached, so `go env` is never invoked for
# targets that don't need it (help, lint, test) nor when GOOS/GOARCH are preset.
GOOS   ?= $(eval GOOS := $(shell $(go_cmd) env GOOS))$(GOOS)
GOARCH ?= $(eval GOARCH := $(shell $(go_cmd) env GOARCH))$(GOARCH)
```

Makefile
mbrandenburger, line 72:
```suggestion
# The per-binary targets are named in the recipe rather than in the prerequisite
# list: prerequisites are expanded when the makefile is parsed, which would
# resolve GOOS/GOARCH on every make invocation.
.PHONY: release-bins
release-bins: ## Cross-compiles all tools for $(GOOS)/$(GOARCH) into $(RELEASE_DIR)
     @$(MAKE) --no-print-directory GOOS=$(GOOS) GOARCH=$(GOARCH) \
             $(TOOLS_EXES:%=$(RELEASE_BIN_DIR)/%)
```

mbrandenburger (commented):
Thank you @pasquale95 for this PR! Great stuff! Please see my comments below.

Makefile
pasquale95: `idemixgen` is not part of this repo. We embed it in `fabric-x-tools` just for convenience because there's no proper image in the idemixgen project itself. Nevertheless, `idemixgen` can be useful in some fabric-x flows.
We may create a binary release workflow in the idemix repo, but this is a separate PR.

pasquale95:
Hi @mbrandenburger, I fixed the PR comments and added a comment about `idemixgen`.

mbrandenburger (approved):
LGTM

PR #302, opened 2026-09-16
add fxadmin utility to image build process

FilipeCapela98:
#### Type of change

- New feature

#### Description

Add `fxadmin` to the `fabric-x-tools` Docker image build so the binary is packaged and shipped alongside the other CLI tools (`configtxgen`, `cryptogen`, `configtxlator`, `fxconfig`).

`fxadmin` was already wired into the Makefile's `TOOLS_EXES`/`pkgmap` for local builds, but the Dockerfile's builder stage never compiled it, so it was missing from the released image.

- Added `RUN go build -o /tmp/bin/fxadmin ./tools/fxadmin` to the Dockerfile builder stage.
- Updated the image's `org.opencontainers.image.description` label to mention `fxadmin`.
- Verified `go build ./tools/fxadmin` succeeds locally.

#### Related issues

N/A


pasquale95 (approved):
LGTM

